PR #22, opened 2023-06-13
Fix finding files with getProjectBase

hamishmack:
The tests use `getProjectBase` to find the `cabal.project` file.  They then look for files based on that.  This change makes that work in nix by copying the `cabal.project` file and required `scripts` file into the parent directory of the component build dir (component build dir will be a subdir `cardano-cli` because we mirror the directory structure of the project for component builds).


newhoggy:
Thanks!

PR #81, opened 2023-07-16
Make it build with ghc-9.6

erikd:
# Changelog

```yaml
- description: Make it build with ghc-9.6
  compatibility: no-api-changes
  type: maintenance
```

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,8 +13,8 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2023-07-05T20:00:33Z
-  , cardano-haskell-packages 2023-07-13T14:00:00Z
+  , hackage.haskell.org 2023-07-10T22:41:49Z
+  , cardano-haskell-packages 2023-07-17T06:35:34Z
 
 packages:
     cardano-cli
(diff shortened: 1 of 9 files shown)


erikd:
~This still has an SRP. Waiting for https://github.com/input-output-hk/cardano-api/pull/122 to be merged.~

SRP has been removed.

cabal.project
```diff
@@ -37,6 +37,19 @@ test-show-details: direct
 -- Always write GHC env files, because they are needed for ghci.
 write-ghc-environment-files: always
 
+-- https://github.com/obsidiansystems/dependent-sum-template/issues/5
+if impl(ghc >= 9.2)
```
andreabedini: Does this mean this is necessary also for ghc-9.4?

andreabedini:
Flake is busted because of [input-output-hk/haskell.nix#](https://github.com/input-output-hk/haskell.nix/pull/2003).

To push this forward, I am changing the lock file to just before the broken commit in haskell.nix.

Edit: PR merged, bumped haskell.nix input again.

carbolymer (approved):

cabal.project
```diff
@@ -37,6 +37,19 @@ test-show-details: direct
...
+  constraints :
+      dependent-sum-template < 0.1.2
```
michaelpj: Why is this necessary in the build? Please leave a comment otherwise we'll never know when we can remove it.

I know we had a problem here because `plutus-core` pulled in the newer `dependent-sum-template` but I think we can fix that much more nicely by revising `plutus-core` in CHaP (which I've done) rather than propagating such conditionals everywhere.

andreabedini: I think it is tricky to understand what can be fixed and where it should be fixed. Maybe some tooling or simply documentation could help? E.g. how could Erik realise this could have been fixed in plutus-core?

erikd: It was a dependency from (I think) plutus-core.

erikd: I was actually able to remove that conditional.

michaelpj: @andreabedini yeah, I think this is not so easy with plain cabal, I don't think it gives you the stack of things that needed the library that doesn't build. I guess I usually get that info from Nix :shrug: 

Possibly building with `--constraint dependent-sum-template < 0` will get the solver to tell you what requires it.

I know I keep harping on about this, but it is _so_ much easier to maintain this stuff if we can fix bounds in CHaP once and for all rather than having a thin spread of conditional logic across our many project files.

PR #162, opened 2023-08-10
Use parameterised by era `ProtocolParameters` type from `cardano-api`

newhoggy:
# Changelog

```yaml
- description: |
  type:
```

# Context

Additional context for the PR goes here.

If the PR fixes a particular issue please provide a
[link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=)
to the issue.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
@@ -234,8 +234,8 @@ runQueryConstitutionHash socketPath (AnyConsensusModeParams cModeParams) network
           handleIOExceptT (ShelleyQueryCmdWriteFileError . FileIOError fpath) $
             LBS.writeFile fpath (encodePretty cHash)
 
-runQueryProtocolParameters
-  :: SocketPath
+runQueryProtocolParameters :: ()
+  => SocketPath
   -> AnyConsensusModeParams
   -> NetworkId
   -> Maybe (File () Out)
@@ -255,18 +255,20 @@ runQueryProtocolParameters socketPath (AnyConsensusModeParams cModeParams) netwo
     eInMode <- toEraInMode era cMode
       & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
 
-    lift (queryProtocolParameters eInMode sbe)
+    result <- lift (queryProtocolParameters eInMode sbe)
       & onLeft (left . ShelleyQueryCmdUnsupportedNtcVersion)
       & onLeft (left . ShelleyQueryCmdEraMismatch)
+
+    pure (InAnyEra era result)
 
   writeProtocolParameters mOutFile =<< except (join (first ShelleyQueryCmdAcquireFailure result))
 
  where
   writeProtocolParameters
     :: Maybe (File () Out)
-    -> ProtocolParameters
+    -> InAnyEra era ProtocolParameters
     -> ExceptT ShelleyQueryCmdError IO ()
-  writeProtocolParameters mOutFile' pparams =
+  writeProtocolParameters mOutFile' (InAnyEra _ pparams) =
     case mOutFile' of
       Nothing -> liftIO $ LBS.putStrLn (encodePretty pparams)
       Just (File fpath) ->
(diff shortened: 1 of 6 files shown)


cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
```diff
@@ -255,18 +255,20 @@ runQueryProtocolParameters socketPath (AnyConsensusModeParams cModeParams) netwo
     eInMode <- toEraInMode era cMode
       & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
 
-    lift (queryProtocolParameters eInMode sbe)
+    result <- lift (queryProtocolParameters eInMode sbe)
       & onLeft (left . ShelleyQueryCmdUnsupportedNtcVersion)
       & onLeft (left . ShelleyQueryCmdEraMismatch)
 
+    pure (InAnyEra era result)
```
newhoggy, line 262: Here, it is not possible to return `result` from this computation because it is parameterised by `era` and `era` is not allowed to escape this scope.  By putting `result` in `InAnyEra` it can be smuggled out of this context and used later in `writeProtocolParameters`.

cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
```diff
@@ -255,18 +255,20 @@ runQueryProtocolParameters socketPath (AnyConsensusModeParams cModeParams) netwo
...
   writeProtocolParameters mOutFile =<< except (join (first ShelleyQueryCmdAcquireFailure result))
 
  where
   writeProtocolParameters
     :: Maybe (File () Out)
-    -> ProtocolParameters
+    -> InAnyEra era ProtocolParameters
     -> ExceptT ShelleyQueryCmdError IO ()
-  writeProtocolParameters mOutFile' pparams =
+  writeProtocolParameters mOutFile' (InAnyEra _ pparams) =
```
newhoggy, line 271: Here, even though we are outside the scope of the computation that produced the `pparams`, we are able to have access to the same `era` type that was available within that computation's context.

cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
newhoggy: See https://github.com/input-output-hk/cardano-cli/pull/162/files#r1290243998

cardano-cli/src/Cardano/CLI/Run/Legacy/Transaction.hs
```diff
@@ -286,8 +286,8 @@ runTransactionCmds cmd =
       runTxSign txinfile skfiles network txoutfile
     TxSubmit mNodeSocketPath anyConsensusModeParams network txFp ->
       runTxSubmit mNodeSocketPath anyConsensusModeParams network txFp
-    TxCalculateMinFee txbody nw pParamsFile nInputs nOutputs nShelleyKeyWitnesses nByronKeyWitnesses ->
-      runTxCalculateMinFee txbody nw pParamsFile nInputs nOutputs nShelleyKeyWitnesses nByronKeyWitnesses
+    TxCalculateMinFee anyEra txbody nw pParamsFile nInputs nOutputs nShelleyKeyWitnesses nByronKeyWitnesses ->
+      runTxCalculateMinFee anyEra txbody nw pParamsFile nInputs nOutputs nShelleyKeyWitnesses nByronKeyWitnesses
```
newhoggy, line 290: Calculating the min fee will now require the era to be specified.  This makes sense because the calculation can be era-sensitive.

cardano-cli/src/Cardano/CLI/Run/Legacy/Transaction.hs
```diff
@@ -1189,24 +1192,28 @@ runTxSubmit socketPath (AnyConsensusModeParams cModeParams) network txFilePath =
...
+      (TxInCount nInputs) (TxOutCount nOutputs)
+      (TxShelleyWitnessCount nShelleyKeyWitnesses)
+      (TxByronWitnessCount nByronKeyWitnesses) = do
+    AnyCardanoEra era <- pure anyEra
 
     txbodyFile <- liftIO $ fileOrPipe txbodyFilePath
     unwitnessed <- firstExceptT ShelleyTxCmdCddlError . newExceptT
                      $ readFileTxBody txbodyFile
-    pparams <- firstExceptT ShelleyTxCmdProtocolParamsError $ readProtocolParameters pParamsFile
+    pparams <- firstExceptT ShelleyTxCmdProtocolParamsError $ readProtocolParameters era pParamsFile
```
newhoggy, line 1216: The reason the `era` is necessary is because the process of reading the protocol parameters which happens inside this function is era-sensitive.

cardano-cli/src/Cardano/CLI/Run/Legacy/Query.hs
```diff
@@ -457,13 +459,16 @@ runQueryKesPeriodInfo socketPath (AnyConsensusModeParams cModeParams) network no
...
+              & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
+
             -- We check that the KES period specified in the operational certificate is correct
             -- based on the KES period defined in the genesis parameters and the current slot number
             eraInMode <- calcEraInMode era $ consensusModeOnly cModeParams
 
             requireNotByronEraInByronMode eraInMode
 
-            gParams <- lift (queryGenesisParameters eInMode sbe)
+            gParams <- lift (queryGenesisParameters shelleyInMode ShelleyBasedEraShelley)
```
newhoggy, line 471: Even though we are in a particular `era` we can still query for  the genesis parameters which is Shelley specific.

Moreover multiple queries exist over the same connection.

Something worth considering is renaming `queryGenesisParameters` to `queryShelleyGenesisParameters` and taking only the mode as an argument.

carbolymer (approved):

PR #203, opened 2023-08-24
Read and write `VotingProcedures` files instead of `VotingEntry` files

newhoggy:
# Changelog

```yaml
- description: |
    Read and write `VotingProcedures` files instead of `VotingEntry` files
  type:
```

# Context

Using `VotingProcedures` as the serialisation types allows us to use ledger's serialisation rather than our own.

This PR converts the IO code to use `VotingProcedures`.  The conversion is not complete because `TxBodyContent` still has a `TxVotes` member.

A subsequent PR in `cardano-api` will switch that out.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Commands/Governance.hs
+++ b/cardano-cli/src/Cardano/CLI/Commands/Governance.hs
@@ -7,6 +7,7 @@ import           Cardano.Api
 import qualified Cardano.Api.Ledger as Ledger
 import           Cardano.Api.Shelley
 
+import           Cardano.CLI.Read (singletonVotingProcedures)
 import           Cardano.CLI.Types.Common
 import           Cardano.CLI.Types.Errors.GovernanceCmdError
 import           Cardano.CLI.Types.Governance
@@ -21,8 +22,8 @@ import           Data.Text (Text)
 import qualified Data.Text.Encoding as Text
 import           Data.Word
 
-runGovernanceCreateVoteCmd
-  :: AnyShelleyBasedEra
+runGovernanceCreateVoteCmd :: ()
+  => AnyShelleyBasedEra
   -> Vote
   -> VType
   -> (TxId, Word32)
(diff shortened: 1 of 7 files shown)


cardano-cli/src/Cardano/CLI/Read.hs
```diff
@@ -764,7 +774,77 @@ readVoteFile
...
+mergeVotingProcedures :: ()
+  => VotingProcedures era
+  -> VotingProcedures era
+  -> VotingProcedures era
+mergeVotingProcedures vpsa vpsb =
+  VotingProcedures
+    $ Ledger.VotingProcedures
+    $ Map.unionWith (Map.unionWith const)
+        (Ledger.unVotingProcedures (unVotingProcedures vpsa))
+        (Ledger.unVotingProcedures (unVotingProcedures vpsb))
```
newhoggy, line 805: The above three functions can be moved into `cardano-api`.  Having it here proves it works and is useful.

cardano-cli/src/Cardano/CLI/Read.hs
```diff
@@ -764,7 +774,77 @@ readVoteFile
...
+  => ShelleyBasedEra era
+  -> Ledger.Voter (Ledger.EraCrypto (ShelleyLedgerEra era))
+  -> Voter era
+fromVoterRole _ = \case
+  Ledger.CommitteeVoter cred ->
+    VoterCommittee (VotingCredential (Ledger.coerceKeyRole cred))    -- TODO: Conway era - We shouldn't be using coerceKeyRole.
+  Ledger.DRepVoter cred ->
+    VoterDRep (VotingCredential cred)
+  Ledger.StakePoolVoter kh ->
+    VoterSpo (StakePoolKeyHash kh)
```
newhoggy, line 818: This can be deleted when `votingProceduresToTxVotes` is deleted.

cardano-cli/src/Cardano/CLI/Commands/Governance.hs
```diff
@@ -31,41 +32,34 @@ runGovernanceCreateVoteCmd
   -> ExceptT GovernanceCmdError IO ()
 runGovernanceCreateVoteCmd anyEra vChoice vType (govActionTxId, govActionIndex) votingStakeCred oFp = do
```
Jimbo4350, line 33: We should deprecate this command (in a separate PR) as we are maintaining two vote creation commands and this one doesn't allow vote creation for DReps or committee members.

cardano-cli/src/Cardano/CLI/Commands/Governance.hs
```diff
@@ -31,41 +32,34 @@ runGovernanceCreateVoteCmd
...
-                                    , votingEntryGovActionId = GovernanceActionId govActIdentifier
-                                    , votingEntryVotingProcedure = voteProcedure
-                                    }
-      firstExceptT WriteFileError . newExceptT $ shelleyBasedEraConstraints sbe $ writeFileTextEnvelope oFp Nothing votingEntry
+      VDR -> do
+        votingCred <- hoistEither $ first VotingCredentialDecodeGovCmdEror $ toVotingCredential sbe vStakeCred
+        let voter = Ledger.DRepVoter (unVotingCredential votingCred)
+            govActIdentifier = createGovernanceActionId govActionTxId govActionIndex
+            voteProcedure = createVotingProcedure sbe vChoice Nothing
+            votingProcedures = singletonVotingProcedures sbe voter govActIdentifier (unVotingProcedure voteProcedure)
```
Jimbo4350, line 54: Not for this PR but there is no reason why we shouldn't allow multiple voting files to be added to a tx.

cardano-cli/src/Cardano/CLI/Read.hs
Jimbo4350: We can leave them here in this PR but we should move them to `cardano-api` so others can use them without depending on `cardano-cli`.

cardano-cli/src/Cardano/CLI/Read.hs
Jimbo4350: This can be deleted in a follow up PR :+1:

Jimbo4350 (approved):
LGTM! :+1:

PR #230, opened 2023-09-01
Era sensitive transaction run commands

newhoggy:
# Changelog

```yaml
- description: |
    Era sensitive transaction run commands
  type:
```

# Context

Era-based versions of transaction commands

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Commands.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Commands.hs
@@ -13,8 +13,10 @@ module Cardano.CLI.EraBased.Commands
 import           Cardano.Api (CardanoEra (..), ShelleyBasedEra (..))
 
 import           Cardano.CLI.Environment
+import           Cardano.CLI.EraBased.Commands.Transaction
 import           Cardano.CLI.EraBased.Options.Common
 import           Cardano.CLI.EraBased.Options.Governance
+import           Cardano.CLI.EraBased.Options.Transaction
 
 import           Data.Foldable
 import           Data.Text (Text)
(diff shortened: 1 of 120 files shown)


carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Commands.hs
```diff
@@ -28,12 +30,14 @@ renderAnyEraCommand :: AnyEraCommand -> Text
 renderAnyEraCommand = \case
   AnyEraCommandOf _ cmd -> renderEraBasedCommand cmd
 
-newtype EraBasedCommand era
+data EraBasedCommand era
   = EraBasedGovernanceCmds (EraBasedGovernanceCmds era)
+  | TransactionCmds (TransactionCmds era)
```
Jimbo4350, line 35: Do we also want to name this `EraBasedTransactionCmds`?

Jimbo4350 (approved):
LGTM. This doesn't break the existing tx build commands right?

cardano-cli/src/Cardano/CLI/EraBased/Commands.hs
newhoggy: This doesn't break existing commands.

My thoughts are we don't need to have the `EraBased` prefix because we already qualify the legacy stuff with `Legacy` and when we get to the point where we can delete the legacy stuff `EraBased` is the only state of things and no prefix is necessary to distinguish it from anything else.

I plan to remove `EraBased` from prefix from the governance commands.

newhoggy: I plan to remove `EraBased` from prefix from the governance commands.

PR #279, opened 2023-09-14
Regularise era based command structure

newhoggy:
# Changelog

```yaml
- description: |
    Regularise era based command structure
  type:
```

# Context

The main change in this PR is to change command parsers to return `Maybe (Parser a)` instead of `Parser a`.

This is based on the principle that commands themselves should determined whether they exist in a given era.  This principle is used in the new Conway commands and will see more use in the future.

Requiring that all command groups accept `Maybe (Parser a)` will mean that any new or existing commands can become era-sensitive without inducing refactoring.  As it is at the moment if a command needs to become era-sensitive, the parent command group will need to be refactored to accomodate it if it doesn't already which is the case for a number top-level commands groups.

The PR also contains a small number of help text fixes.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
@@ -24,7 +24,7 @@ pAddressCmds _ envCli =
   subInfoParser "address"
     ( Opt.progDesc
         $ mconcat
-          [ "Address commands."
+          [ "Payment address commands."
           ]
     )
     [ Just
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
@@ -13,14 +13,20 @@ import           Cardano.CLI.Environment (EnvCli (..))
 import           Cardano.CLI.EraBased.Commands.Address
 import           Cardano.CLI.EraBased.Options.Common
 
-import           Data.Foldable
-import           Data.Maybe
 import           Options.Applicative hiding (help, str)
 import qualified Options.Applicative as Opt
 
-pAddressCmds :: CardanoEra era -> EnvCli -> Parser (AddressCmds era)
+pAddressCmds :: ()
+  => CardanoEra era
+  -> EnvCli
+  -> Maybe (Parser (AddressCmds era))
 pAddressCmds _ envCli =
-  asum $ catMaybes
+  subInfoParser "address"
+    ( Opt.progDesc
+        $ mconcat
+          [ "Address commands."
+          ]
+    )
     [ Just
         $ subParser "key-gen"
         $ Opt.info pAddressKeyGen
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
@@ -14,22 +14,27 @@ import           Cardano.CLI.EraBased.Commands.Address
 import           Cardano.CLI.EraBased.Options.Common
 
 import           Data.Foldable
+import           Data.Maybe
 import           Options.Applicative hiding (help, str)
 import qualified Options.Applicative as Opt
 
 pAddressCmds :: CardanoEra era -> EnvCli -> Parser (AddressCmds era)
 pAddressCmds _ envCli =
-  asum
-    [ subParser "key-gen"
+  asum $ catMaybes
+    [ Just
+        $ subParser "key-gen"
         $ Opt.info pAddressKeyGen
         $ Opt.progDesc "Create an address key pair."
-    , subParser "key-hash"
+    , Just
+        $ subParser "key-hash"
         $ Opt.info pAddressKeyHash
         $ Opt.progDesc "Print the hash of an address key."
-    , subParser "build"
+    , Just
+        $ subParser "build"
         $ Opt.info (pAddressBuild envCli)
         $ Opt.progDesc "Build a Shelley payment address, with optional delegation to a stake address."
-    , subParser "info"
+    , Just
+        $ subParser "info"
         $ Opt.info pAddressInfo
         $ Opt.progDesc "Print information about an address."
     ]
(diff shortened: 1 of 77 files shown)


carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
```diff
@@ -14,22 +14,27 @@ import           Cardano.CLI.EraBased.Commands.Address
...
 import           Data.Foldable
+import           Data.Maybe
 import           Options.Applicative hiding (help, str)
 import qualified Options.Applicative as Opt
 
 pAddressCmds :: CardanoEra era -> EnvCli -> Parser (AddressCmds era)
 pAddressCmds _ envCli =
-  asum
-    [ subParser "key-gen"
+  asum $ catMaybes
```
Jimbo4350: I don't think this is necessary as these commands will be available in every era unless e.g the crypto changes.

cardano-cli/src/Cardano/CLI/EraBased/Options/Genesis.hs
```diff
@@ -27,50 +27,61 @@ import qualified Options.Applicative as Opt
 
 pGenesisCmds :: EnvCli -> Parser (GenesisCmds era)
 pGenesisCmds envCli =
-  asum
-    [ subParser "key-gen-genesis"
+  asum $ catMaybes
```
Jimbo4350: I'm not sure I understand the reason for this change.

cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
```diff
@@ -13,23 +13,34 @@ import           Cardano.CLI.Environment (EnvCli (..))
...
-import           Data.Foldable
 import           Options.Applicative hiding (help, str)
 import qualified Options.Applicative as Opt
 
-pAddressCmds :: CardanoEra era -> EnvCli -> Parser (AddressCmds era)
+pAddressCmds :: ()
+  => CardanoEra era
+  -> EnvCli
+  -> Maybe (Parser (AddressCmds era))
 pAddressCmds _ envCli =
```
Jimbo4350, line 23: So we're not using the `CardanoEra` value here because these commands are constant for all the eras. I'm assuming you're doing this in the event that e.g the cryptography changes in a future era?

cardano-cli/src/Cardano/CLI/EraBased/Options/Genesis.hs
```diff
@@ -16,61 +16,78 @@ import           Cardano.CLI.EraBased.Options.Common
...
 pGenesisCmds envCli =
-  asum
-    [ subParser "key-gen-genesis"
+  subInfoParser "genesis"
+    ( Opt.progDesc
+        $ mconcat
+          [ "Genesis block commands."
+          ]
+    )
+    [ Just
```
Jimbo4350, line 37: Why the change from `asum`?

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Query.hs
```diff
@@ -14,37 +14,36 @@ import           Cardano.CLI.EraBased.Options.Common
...
 
-pGovernanceQueryGetConstitutionCmd
-  :: EnvCli
-  -> CardanoEra era
+pGovernanceQueryGetConstitutionCmd :: ()
+  => CardanoEra era
+  -> EnvCli
   -> Maybe (Parser (GovernanceQueryCmds era))
-pGovernanceQueryGetConstitutionCmd env era = do
+pGovernanceQueryGetConstitutionCmd era env = do
```
Jimbo4350, line 35: :+1:

cardano-cli/src/Cardano/CLI/EraBased/Options/Key.hs
```diff
@@ -21,65 +21,78 @@ import qualified Options.Applicative as Opt
 {- HLINT ignore "Use <$>" -}
 {- HLINT ignore "Move brackets to avoid $" -}
 
-pKeyCmds :: Parser (KeyCmds era)
+pKeyCmds :: Maybe (Parser (KeyCmds era))
 pKeyCmds =
-  asum
```
Jimbo4350, line 26: Why the change from `Parser (..)` -> `Maybe (Parser (...))`

Jimbo4350 (dismissed):
Some questions

cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
newhoggy: Or if we add new commands that are only apply to a future era.

cardano-cli/src/Cardano/CLI/EraBased/Options/Key.hs
newhoggy: The change represents a shift of responsibilities.

Previously it was the parent command group that decides whether the command should appear in an era.

The change shifts the responsibility from the parent to the command in question.  ie the command itself determines whether or not it should show up in a given era.

This is already how it is done for the governance commands and some other commands.

The purpose of this PR is to harmonise our code base to consistently use the latter convention.

cardano-cli/src/Cardano/CLI/EraBased/Options/Address.hs
newhoggy: Some scenarios that might cause the era to be need:

* A new command or flag added that only apply to newer eras
* An existing command or flag needs to be deleted for newer eras

cardano-cli/src/Cardano/CLI/EraBased/Options/Genesis.hs
newhoggy: `asum` is in the implementation of `subInfoParser` so we are still calling `asum` indirectly:

```
subInfoParser :: String -> InfoMod a -> [Maybe (Parser a)] -> Maybe (Parser a)
subInfoParser name i mps = case catMaybes mps of
  [] -> Nothing
  parsers -> Just $ subParser name $ Opt.info (asum parsers) i
```

`subInfoParser` is used instead so that commands are self-describing.  Previously the implementation of the parent command group provided the description.

PR #322, opened 2023-10-02
create-poll, answer-poll, verify-poll: move to 'babbage governance' block

smelc:
# Changelog

```yaml
- description: |
    create-poll, answer-poll, verify-poll: move to 'babbage governance' block
  type:
```

# Context

Part of fixing https://github.com/input-output-hk/cardano-cli/issues/313

# Checklist

- [X] The change log section in the PR description has been filled in
- [X] New tests are added if needed and existing tests are updated.  These may include:
  - golden tests
  - property tests
  - round trip tests
  - integration tests
- NA The version bounds in `.cabal` files are updated
- [X] CI passes. See note on CI.  The following CI checks are required:
  - [X] Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` to get the `hlint` version
  - [X] Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version
  - [X] Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`
- [X] Self-reviewed the diff


cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-module Test.Golden.Shelley.Governance.AnswerPoll where
+module Test.Golden.Babbage.Governance.AnswerPoll where
```
newhoggy: The `Shelley` in `Test.Golden.Shelley.Governance.AnswerPoll` doesn't mean `Shelley` era, it means "Shelley Based Era", which is confusing.

When in era-based, we shouldn't organise by era because it can get complicated.  We should omit both era and eon.

Maybe the module should be`Test.Golden.EraBased.Governance.AnswerPoll` to say we are testing era-based commands.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/VerifyPoll.hs
```diff
@@ -19,15 +19,15 @@ import qualified Hedgehog.Internal.Property as H
...
 {- HLINT ignore "Use camelCase" -}
 
-hprop_golden_shelleyGovernanceVerifyPoll :: Property
-hprop_golden_shelleyGovernanceVerifyPoll = propertyOnce $ do
-  pollFile <- noteInputFile "test/cardano-cli-golden/files/golden/shelley/governance/polls/basic.json"
-  txFile <- noteInputFile "test/cardano-cli-golden/files/golden/shelley/governance/verify/valid"
+hprop_golden_babbageGovernanceVerifyPoll :: Property
+hprop_golden_babbageGovernanceVerifyPoll = propertyOnce $ do
+  pollFile <- noteInputFile "test/cardano-cli-golden/files/golden/babbage/governance/polls/basic.json"
+  txFile <- noteInputFile "test/cardano-cli-golden/files/golden/babbage/governance/verify/valid"
```
newhoggy: Don't include `babbage` as a top-level directory.

If you must include the era in the name because you are testing multiple eras (which isn't the case here), put the era at the lowest level in the filename instead.

This might see counter-intuitive because our commands have era at the top-level, but actually the era gets passed to the command as an argument so operationally, it's more like run this command, and here is the era the command should run in.

Moreover, it is often the case commands behave the same in multiple eras so it isn't necessary to qualify by era.

In the above example, this ought to be sufficient:

* `test/cardano-cli-golden/files/golden/governance/polls/basic.json`
* `test/cardano-cli-golden/files/golden/governance/verify/valid`

We will never run these commands outside of babbage so any babbage qualification is unnecessary.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
```diff
@@ -30,43 +30,43 @@ hprop_golden_shelleyGovernanceAnswerPollNeg1Invalid = propertyOnce . H.moduleWor
 
   either (const H.success) (const H.failure) result
 
-hprop_golden_shelleyGovernanceAnswerPoll0 :: Property
-hprop_golden_shelleyGovernanceAnswerPoll0 = propertyOnce . H.moduleWorkspace "governance-answer-poll" $ \tempDir -> do
-  pollFile <- noteInputFile "test/cardano-cli-golden/files/golden/shelley/governance/polls/basic.json"
-  expectedAnswerFile <- noteInputFile "test/cardano-cli-golden/files/golden/shelley/governance/polls/basic.answer.0.json"
+hprop_golden_babbageGovernanceAnswerPoll0 :: Property
+hprop_golden_babbageGovernanceAnswerPoll0 = propertyOnce . H.moduleWorkspace "governance-answer-poll" $ \tempDir -> do
+  pollFile <- noteInputFile "test/cardano-cli-golden/files/golden/babbage/governance/polls/basic.json"
```
newhoggy: This should be `test/cardano-cli-golden/files/input/governance/polls/basic.json`

That is:

* Replace `golden` with `input` because this is not a golden file.
* Delete `babbage`.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
```diff
@@ -30,43 +30,43 @@ hprop_golden_shelleyGovernanceAnswerPollNeg1Invalid = propertyOnce . H.moduleWor
...
+  expectedAnswerFile <- noteInputFile "test/cardano-cli-golden/files/golden/babbage/governance/polls/basic.answer.0.json"
```
newhoggy: This is a golden file, so the variable name should say `goldenAnswerFile`.

Golden files are not input files, so use `note` instead of `noteInputFile`.  (In future, there should be a `noteGoldenFile` function)

Also, drop `babbage`.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
smelc: OK I see :+1:

I changed to `Test.Golden.EraBased.Governance.AnswerPoll`.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/VerifyPoll.hs
smelc: Thanks for the explanations :pray: I removed the `babbage` segment from paths.

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
smelc: I did the change :+1:

cardano-cli/test/cardano-cli-golden/Test/Golden/Babbage/Governance/AnswerPoll.hs
smelc: Changed the names :+1: Thanks for the suggestions.

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Poll.hs
```diff
@@ -0,0 +1,37 @@
...
+      [Text] -- ^ Choices
+      (Maybe Word) -- ^ Nonce
+      (File GovernancePoll Out)
+  | GovernanceAnswerPoll  -- ^ Answer a SPO poll
+      (BabbageEraOnwards era) {- TODO smelc, use BabbageEraOnly here instead -}
+      (File GovernancePoll In) -- ^ Poll file
+      (Maybe Word) -- ^ Answer index
+      (Maybe (File () Out)) -- ^ Tx file
+  | GovernanceVerifyPoll  -- ^ Verify answer to a given SPO poll
+      (BabbageEraOnwards era) {- TODO smelc, use BabbageEraOnly here instead -}
```
newhoggy, line 25: Are these TODOs still relevant?

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
```diff
@@ -0,0 +1,74 @@
...
+
+import           Control.Monad (when)
+import           Data.Foldable
+import           Options.Applicative hiding (help, str)
+import qualified Options.Applicative as Opt
+
+pGovernancePollCmds ::
+  () =>
+  CardanoEra era ->
+  Maybe (Parser (GovernancePollCmds era))
```
newhoggy: Can we have the arrows on the LHS on the next line?

newhoggy (approved):

newhoggy:
A few more comments.  Merge when you're ready.

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Poll.hs
smelc: Yes they are, because the `poll` commands are babbage only, whereas this is babbage onwards.

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
smelc: :+1: Did the change:

```diff
-pGovernancePollCmds ::
-  () =>
-  CardanoEra era ->
-  Maybe (Parser (GovernancePollCmds era))
+pGovernancePollCmds :: ()
+  => CardanoEra era
+  -> Maybe (Parser (GovernancePollCmds era))
```

cardano-cli/test/cardano-cli-golden/files/golden/help.cli
```diff
@@ -8777,25 +8824,6 @@ Usage: cardano-cli legacy governance create-update-proposal --out-file FILE
...
-                                                   [--out-file FILE]
-
-  Answer an SPO poll
-
-Usage: cardano-cli legacy governance verify-poll --poll-file FILE
-                                                   --tx-file FILE
-                                                   [--out-file FILE]
-
-  Verify an answer to a given SPO poll
-
```
newhoggy, line 8798: Actually, an oversight in my review.  We're not actually supposed to delete the legacy commands.  These need to be kept in tact.  We will eventually delete all the legacy commands in one go, but not right now.

PR #349, opened 2023-10-09
Plug *-poll legacy commands back

### LINKED ISSUE #340: Plug *-poll legacy commands back
Actually, an oversight in my review.  We're not actually supposed to delete the legacy commands.  These need to be kept in tact.  We will eventually delete all the legacy commands in one go, but not right now.

_Originally posted by @newhoggy in https://github.com/input-output-hk/cardano-cli/pull/322#discussion_r1347198164_

smelc:
# Changelog

```yaml
- description: |
    Bring back legacy *-poll commands
  type:
```

# Context

@newhoggy witnessed after https://github.com/input-output-hk/cardano-cli/pull/322's merge that it was removing two many cmomands: the `legacy` voting commands should have been kept.

This PR brings back the `legacy` poll commands.

Fix https://github.com/input-output-hk/cardano-cli/issues/340

# Checklist

- [X] The change log section in the PR description has been filled in
- NA New tests are added if needed and existing tests are updated.  These may include:
  - golden tests
  - property tests
  - round trip tests
  - integration tests
- NA The version bounds in `.cabal` files are updated
- [X] CI passes. See note on CI.  The following CI checks are required:
  - [X] Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` to get the `hlint` version
  - [X] Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version
  - [X] Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`
- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
@@ -4,7 +4,10 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.CLI.EraBased.Run.Governance.Poll
-  ( runGovernancePollCmds
+  ( runGovernancePollCmds,
+    runGovernanceCreatePoll,
+    runGovernanceAnswerPoll,
+    runGovernanceVerifyPoll
   ) where
 
 import           Cardano.Api
--- a/cardano-cli/src/Cardano/CLI/Legacy/Commands/Governance.hs
+++ b/cardano-cli/src/Cardano/CLI/Legacy/Commands/Governance.hs
@@ -35,6 +35,19 @@ data LegacyGovernanceCmds
       [VerificationKeyFile In]
       ProtocolParametersUpdate
       (Maybe FilePath)
+  | GovernanceCreatePoll
+      Text -- ^ Prompt
+      [Text] -- ^ Choices
+      (Maybe Word) -- ^ Nonce
+      (File GovernancePoll Out)
+  | GovernanceAnswerPoll
+      (File GovernancePoll In) -- ^ Poll file
+      (Maybe Word) -- ^ Answer index
+      (Maybe (File () Out)) -- ^ Tx file
+  | GovernanceVerifyPoll
+      (File GovernancePoll In) -- ^ Poll file
+      (File (Tx ()) In) -- ^ Tx file
+      (Maybe (File () Out)) -- ^ Tx file
   deriving Show
 
 renderLegacyGovernanceCmds :: LegacyGovernanceCmds -> Text
@@ -44,4 +57,7 @@ renderLegacyGovernanceCmds = \case
   GovernanceMIRTransfer _ _ _ TransferToTreasury -> "governance create-mir-certificate transfer-to-treasury"
   GovernanceMIRTransfer _ _ _ TransferToReserves -> "governance create-mir-certificate transfer-to-reserves"
   GovernanceUpdateProposal {} -> "governance create-update-proposal"
+  GovernanceCreatePoll{} -> "governance create-poll"
+  GovernanceAnswerPoll{} -> "governance answer-poll"
+  GovernanceVerifyPoll{} -> "governance verify-poll"
 
--- a/cardano-cli/src/Cardano/CLI/Legacy/Options.hs
+++ b/cardano-cli/src/Cardano/CLI/Legacy/Options.hs
@@ -1032,6 +1032,15 @@ pGovernanceCmds envCli =
     , subParser "create-update-proposal"
         $ Opt.info pUpdateProposal
         $ Opt.progDesc "Create an update proposal"
+    , subParser "create-poll"
+        $ Opt.info pGovernanceCreatePoll
+        $ Opt.progDesc "Create an SPO poll"
+    , subParser "answer-poll"
+        $ Opt.info pGovernanceAnswerPoll
+        $ Opt.progDesc "Answer an SPO poll"
+    , subParser "verify-poll"
+        $ Opt.info pGovernanceVerifyPoll
+        $ Opt.progDesc "Verify an answer to a given SPO poll"
     ]
   where
     mirCertParsers :: Parser LegacyGovernanceCmds
@@ -1089,6 +1098,28 @@ pGovernanceCmds envCli =
         <*> some pGenesisVerificationKeyFile
         <*> pProtocolParametersUpdate
         <*> optional pCostModels
+
+    pGovernanceCreatePoll :: Parser LegacyGovernanceCmds
+    pGovernanceCreatePoll =
+      GovernanceCreatePoll
+        <$> pPollQuestion
+        <*> some pPollAnswer
+        <*> optional pPollNonce
+        <*> pOutputFile
+
+    pGovernanceAnswerPoll :: Parser LegacyGovernanceCmds
+    pGovernanceAnswerPoll =
+      GovernanceAnswerPoll
+        <$> pPollFile
+        <*> optional pPollAnswerIndex
+        <*> optional pOutputFile
+
+    pGovernanceVerifyPoll :: Parser LegacyGovernanceCmds
+    pGovernanceVerifyPoll =
+      GovernanceVerifyPoll
+        <$> pPollFile
+        <*> pPollTxFile
+        <*> optional pOutputFile
 
 pGenesisCmds :: EnvCli -> Parser LegacyGenesisCmds
 pGenesisCmds envCli =
--- a/cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
+++ b/cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
@@ -24,6 +24,7 @@ import           Data.Aeson (eitherDecode)
 import qualified Data.ByteString.Lazy as LB
 import           Data.Function ((&))
 import qualified Data.Text as Text
+import Cardano.CLI.EraBased.Run.Governance.Poll
 
 runLegacyGovernanceCmds :: LegacyGovernanceCmds -> ExceptT GovernanceCmdError IO ()
 runLegacyGovernanceCmds = \case
@@ -35,6 +36,13 @@ runLegacyGovernanceCmds = \case
     runLegacyGovernanceGenesisKeyDelegationCertificate sbe genVk genDelegVk vrfVk out
   GovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp ->
     runLegacyGovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp
+  GovernanceCreatePoll prompt choices nonce out ->
+    runGovernanceCreatePoll BabbageEraOnwardsBabbage  prompt choices nonce out
+  GovernanceAnswerPoll poll ix mOutFile ->
+    runGovernanceAnswerPoll BabbageEraOnwardsBabbage poll ix mOutFile
+  GovernanceVerifyPoll poll metadata mOutFile ->
+    runGovernanceVerifyPoll BabbageEraOnwardsBabbage poll metadata mOutFile
+
 
 runLegacyGovernanceMIRCertificatePayStakeAddrs
   :: EraInEon ShelleyToBabbageEra
--- a/cardano-cli/test/cardano-cli-golden/files/golden/help.cli
+++ b/cardano-cli/test/cardano-cli-golden/files/golden/help.cli
@@ -8722,6 +8722,9 @@ Usage: cardano-cli legacy governance
                                        ( create-mir-certificate
                                        | create-genesis-key-delegation-certificate
                                        | create-update-proposal
+                                       | create-poll
+                                       | answer-poll
+                                       | verify-poll
                                        )
 
   Governance commands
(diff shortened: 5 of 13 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance.hs
```diff
@@ -38,7 +38,7 @@ data GovernanceCmds era
   | GovernanceDRepCmds
       (GovernanceDRepCmds era)
   | GovernancePollCmds
-      (GovernancePollCmds era)
+      GovernancePollCmds {- TODO Add era when legacy commands are removed -}
```
newhoggy, line 41: I don't think we need to do this if the `era` type argument is not used.

newhoggy: The `era` type argument should stay so this command will only appear in the Babbage era.

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Poll.hs
```diff
@@ -9,27 +9,28 @@ import           Cardano.Api.Shelley
 
 import           Data.Text (Text)
 
-data GovernancePollCmds era
+data GovernancePollCmds
   = GovernanceCreatePoll -- ^ Create a SPO poll
-      (BabbageEraOnwards era) {- TODO smelc, use BabbageEraOnly here instead -}
+      -- (BabbageEraOnwards era) -- TODO Use when legacy command is removed
```
newhoggy: We should keep this and change it to `BabbageEraOnly` as per the original TODO

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
```diff
@@ -1,73 +1,75 @@
...
 import qualified Options.Applicative as Opt
 
+-- | Era-dependent entry point. Only returns @Just _@ if the era
+-- is babbage, since those commands are babbage-specific.
+-- Use the parser of individual commands to avoid providing the era.
+-- See https://github.com/input-output-hk/cardano-cli/issues/340 for context.
 pGovernancePollCmds :: ()
   => CardanoEra era
-  -> Maybe (Parser (GovernancePollCmds era))
+  -> Maybe (Parser GovernancePollCmds)
```
newhoggy: The `era` type argument should be retained.

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
```diff
@@ -1,73 +1,75 @@
...
+            ),
           subParser "verify-poll"
-            <$> ( Opt.info
-                    <$> pGovernanceVerifyPoll era
-                    <*> pure (Opt.progDesc "Verify an answer to a given SPO poll")
-                )
+            ( Opt.info
+                pGovernanceVerifyPoll
+                (Opt.progDesc "Verify an answer to a given SPO poll")
+            )
```
newhoggy: The previous implementation was correct.

cardano-cli/src/Cardano/CLI/Legacy/Options.hs
```diff
@@ -39,6 +39,7 @@ import           Cardano.CLI.Legacy.Commands.TextView
 import           Cardano.CLI.Legacy.Commands.Transaction
 import           Cardano.CLI.Parser
 import           Cardano.CLI.Types.Common
+import           Cardano.CLI.EraBased.Options.Governance.Poll
```
newhoggy: Restore the original legacy parsers instead of using the era-based one.

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
```diff
@@ -35,6 +36,8 @@ runLegacyGovernanceCmds = \case
     runLegacyGovernanceGenesisKeyDelegationCertificate sbe genVk genDelegVk vrfVk out
   GovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp ->
     runLegacyGovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp
+  GovernanceLegacyPoll sub ->
+    runGovernancePollCmds sub
```
newhoggy: We should call `runLegacyGovernancePollCmds`.  The individual legacy poll commands should be the ones delegating to the era-based version.

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
```diff
@@ -32,23 +32,22 @@ import qualified System.IO as IO
 import           System.IO (stderr, stdin, stdout)
 
 
-runGovernancePollCmds :: GovernancePollCmds era -> ExceptT GovernanceCmdError IO ()
+runGovernancePollCmds :: GovernancePollCmds -> ExceptT GovernanceCmdError IO ()
```
newhoggy: We should retain the `era` type argument.

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance.hs
smelc: Done :+1:

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
smelc: Done :+1:

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Poll.hs
smelc: Kept this :+1:

cardano-cli/src/Cardano/CLI/EraBased/Options/Governance/Poll.hs
smelc: Reverted :+1:

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
smelc: Legacy commands now delegate to the era-based version :+1:

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
smelc: Done :+1:

cardano-cli/src/Cardano/CLI/Legacy/Options.hs
smelc: Done :+1:

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
```diff
@@ -34,21 +37,20 @@ import           System.IO (stderr, stdin, stdout)
...
-  GovernanceAnswerPoll w poll ix mOutFile ->
-    runGovernanceAnswerPoll w poll ix mOutFile
-  GovernanceVerifyPoll w poll metadata mOutFile ->
-    runGovernanceVerifyPoll w poll metadata mOutFile
+  GovernanceCreatePoll _ prompt choices nonce out ->
+    runGovernanceCreatePoll prompt choices nonce out
+  GovernanceAnswerPoll _ poll ix mOutFile ->
+    runGovernanceAnswerPoll poll ix mOutFile
+  GovernanceVerifyPoll _ poll metadata mOutFile ->
+    runGovernanceVerifyPoll poll metadata mOutFile
```
newhoggy: Can you pass in the `w` to the run functions as before?

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
```diff
@@ -34,21 +37,20 @@ import           System.IO (stderr, stdin, stdout)
...
 
 runGovernanceCreatePoll
-  :: BabbageEraOnwards era
```
newhoggy: Change this to be `BabbageEraOnly`

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
```diff
@@ -35,6 +36,13 @@ runLegacyGovernanceCmds = \case
     runLegacyGovernanceGenesisKeyDelegationCertificate sbe genVk genDelegVk vrfVk out
   GovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp ->
     runLegacyGovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp
+  GovernanceCreatePoll prompt choices nonce out ->
+    runGovernanceCreatePoll prompt choices nonce out
+  GovernanceAnswerPoll poll ix mOutFile ->
+    runGovernanceAnswerPoll poll ix mOutFile
+  GovernanceVerifyPoll poll metadata mOutFile ->
+    runGovernanceVerifyPoll poll metadata mOutFile
```
newhoggy: Call the era-based run functions using `BabbageEraOnlyBabbage` as the era witness.

smelc: @newhoggy> what is the rule for allowing that? Is that because the caller is from a `Legacy` package and so we are more lenient?

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
smelc: I can't yet, because we need a release of `cardano-api` to bring [the PR](https://github.com/input-output-hk/cardano-api/pull/304) introducing `BabbageEraOnly`. But I will do so when possible :+1:

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
smelc: Done :+1:

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
newhoggy: Good question.  This command is somewhat a special case.  It was introduced during Babbage as a short term feature that will be obsoleted by Conway, hence it is Babbage only.  Even if in the era-based version of the command we want to extend its lifetime to include Conway (which is still a possibility), the legacy commands are deprecated and will not extend to Conway.  Therefore we know for sure the legacy command is Babbage-only.

And because it is Babbage-only, the legacy command can provide a witness that satisfies the eon type and also says it is Babbage.

Most other commands outside of `poll` do not meet this criteria.

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
```diff
@@ -4,7 +4,10 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.CLI.EraBased.Run.Governance.Poll
-  ( runGovernancePollCmds
+  ( runGovernancePollCmds,
+    runGovernanceCreatePoll, {- For legacy caller only -}
+    runGovernanceAnswerPoll, {- For legacy caller only -}
+    runGovernanceVerifyPoll  {- For legacy caller only -}
```
newhoggy: We don't need to add a comment here.  It is okay to export run commands even if no-one uses them.

In fact, it is preferred in case we want to change the tests to use these instead of the CLI.

cardano-cli/src/Cardano/CLI/Legacy/Commands/Governance.hs
```diff
@@ -35,6 +35,19 @@ data LegacyGovernanceCmds
...
+      (Maybe Word) -- Nonce
+      (File GovernancePoll Out)
+  | GovernanceAnswerPoll
+      (File GovernancePoll In) -- Poll file
+      (Maybe Word) -- Answer index
+      (Maybe (File () Out)) -- Tx file
+  | GovernanceVerifyPoll
+      (File GovernancePoll In) -- Poll file
+      (File (Tx ()) In) -- Tx file
+      (Maybe (File () Out)) -- Tx file
```
newhoggy: Comments are good.  If used, then use `-- ^`.

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
```diff
@@ -35,6 +36,16 @@ runLegacyGovernanceCmds = \case
...
   GovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp ->
     runLegacyGovernanceUpdateProposal out eNo genVKeys ppUp mCostModelFp
+  GovernanceCreatePoll prompt choices nonce out ->
+    {- TODO replace BabbageEraOnwardsBabbage by BabbageEraOnlyBabbage when available -}
+    runGovernanceCreatePoll BabbageEraOnwardsBabbage  prompt choices nonce out
+  GovernanceAnswerPoll poll ix mOutFile ->
+    {- TODO replace BabbageEraOnwardsBabbage by BabbageEraOnlyBabbage when available -}
+    runGovernanceAnswerPoll BabbageEraOnwardsBabbage poll ix mOutFile
+  GovernanceVerifyPoll poll metadata mOutFile ->
+    {- TODO replace BabbageEraOnwardsBabbage by BabbageEraOnlyBabbage when available -}
```
newhoggy: We shouldn't have to add these comments because when the `run` command's eon witness type is changed we will get a type error here and know to change it.

In any case, I originally asked for this to be changed to `BabbageEraOnly`.  I now think we should keep the eon type as it is.  When we hard fork to Conway, users of this command should still expect this to continue working.

We can deprecate it post-Conway hard-fork.

newhoggy (approved):

cardano-cli/src/Cardano/CLI/Legacy/Run/Governance.hs
smelc: Removed the comments :+1:

You're right, I add in mind those `poll` commands where `babbage` only, but this is the legacy part we're speaking here; so it doesn't apply.

cardano-cli/src/Cardano/CLI/Legacy/Commands/Governance.hs
smelc: Good catch :+1: Changed.

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Poll.hs
smelc: Changed :+1:

PR #350, opened 2023-10-09
Do not allow submitting transactions older than current node era

newhoggy:
# Changelog

```yaml
- description: |
    Do not allow submitting transactions older than current node era
  type:
```

# Context

Submitting transactions older that the current node era is unsafe.

This PR also removes all uses of `EraCast` from `cardano-api`.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Types/Errors/NodeEraMismatchError.hs
+++ b/cardano-cli/src/Cardano/CLI/Types/Errors/NodeEraMismatchError.hs
@@ -1,0 +1,13 @@
+{-# LANGUAGE GADTs #-}
+
+module Cardano.CLI.Types.Errors.NodeEraMismatchError
+  ( NodeEraMismatchError(..)
+  ) where
+
+import           Cardano.Api
+
+data NodeEraMismatchError = forall era nodeEra.
+  NodeEraMismatchError
+  { era           :: !(CardanoEra era)
+  , nodeEra       :: !(CardanoEra nodeEra)
+  }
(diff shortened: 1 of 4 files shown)


cardano-cli/src/Cardano/CLI/Types/Errors/NodeEraMismatchError.hs
```diff
@@ -0,0 +1,15 @@
...
+module Cardano.CLI.Types.Errors.NodeEraMismatchError
+  ( NodeEraMismatchError(..)
+  ) where
+
+import           Cardano.Api
+
+data NodeEraMismatchError = forall era nodeEra value.
+  Show value =>
+    NodeEraMismatchError
+    { originalValue :: !value
```
carlhammann: What's the purpose of this value?

Jimbo4350: I think we don't need the `originalValue` field. It's probably enough to say the node is in an unexpected era and what era that is.

Jimbo4350 (commented):
One minor comment. Happy to approve after.

cardano-cli/src/Cardano/CLI/Types/Errors/NodeEraMismatchError.hs
newhoggy: The field is so that if we have some value that is in the wrong era for example `TxOut era` then we can print out a useful error message to say what it was that was in the wrong era.  @Jimbo4350 is right that we don't need it because we now outright reject circumstances where eras don't match rather than before when we were more lenient and additional context was helpful.

The field has been removed.

Jimbo4350 (approved):

PR #373, opened 2023-10-13
The `vote-file` and `proposal-file` options are Conway onwards only

newhoggy:
# Changelog

```yaml
- description: |
    The `vote-file` and `proposal-file` options are Conway onwards only
  type:
```

# Context

Use the `Featured` type to declare the `--vote-files` and `--proposal-files` flags in `transaction build` and `transaction build-raw` commands should only exist from Conway era onwards.

Because TxBody also uses `Featured` in the same way, the code that links the CLI types to the TxBody type is streamlined.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -902,12 +902,19 @@ runQueryStakeAddressInfoCmd
 
         requireNotByronEraInByronMode eraInMode
 
-        result <- lift (queryStakeAddresses eInMode sbe stakeAddr networkId)
+        (stakeRewardAccountBalances, stakePools) <- lift (queryStakeAddresses eInMode sbe stakeAddr networkId)
           & onLeft (left . QueryCmdUnsupportedNtcVersion)
           & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
 
-        pure $ do
-          writeStakeAddressInfo mOutFile $ DelegationsAndRewards result
+        stakeDelegDeposits <- lift (queryStakeDelegDeposits eInMode sbe stakeAddr)
+          & onLeft (left . QueryCmdUnsupportedNtcVersion)
+          & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
+
+        return $ do
+          writeStakeAddressInfo
+            mOutFile
+            (DelegationsAndRewards (stakeRewardAccountBalances, stakePools))
+            (Map.mapKeys (makeStakeAddress networkId) stakeDelegDeposits)
     )
     & onLeft (left . QueryCmdAcquireFailure)
     & onLeft left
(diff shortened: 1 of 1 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -167,13 +167,20 @@ runTransactionBuildCmd
 
   -- Conway related
   votingProcedures <-
-    inEonForEra
-      (pure emptyVotingProcedures)
-      (\w -> firstExceptT TxCmdVoteError $ ExceptT (readVotingProceduresFiles w voteFiles))
-      eon
+    case mfVoteFiles of
```
Jimbo4350: So we're now essentially pattern matching twice based on whether or not a feature is in an era.

Jimbo4350 (dismissed):
Sorry I disagree here.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
newhoggy: Can you clarify what you mean?  Is the problem the case match n the `Maybe`, once for voting procedures and once for proposals?

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -165,15 +172,15 @@ runTransactionBuildCmd
 
   txOuts <- mapM (toTxOutInAnyEra eon) txouts
 
-  -- Conway related
   votingProcedures <-
-    inEonForEra
-      (pure emptyVotingProcedures)
-      (\w -> firstExceptT TxCmdVoteError $ ExceptT (readVotingProceduresFiles w voteFiles))
-      eon
```
newhoggy: The use of `inEonForEra` is made unnecessary by using the `Featured` type.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -165,15 +172,15 @@ runTransactionBuildCmd
...
+    liftMaybeFeatured mfVoteFiles $ \w voteFiles ->
+      lift (readVotingProceduresFiles w voteFiles)
+        & onLeft (left . TxCmdVoteError)
```
newhoggy: Using the new `liftMaybeFeatured` function to make it easy to manipulate the value within the `Featured` type.

`readVotingProceduresFiles` takes a `[VoteFile In]` and returns a `VotingProcedures era`

`liftMaybeFeatured` takes that computation to convert a `Maybe (Featured ConwayEraOnwards [VoteFile In])` to a `Maybe (Featured ConwayEraOnwards [VotingProcedures era])`

PR #375, opened 2023-10-13
Make `stake-address-info` query also return deposit

### LINKED ISSUE #353: [FEATUE] A query for the stake key deposited amount | Stake-Address-Info is not showing the deposited amount
Version Info:
```
cardano-cli 8.12.0.0 - linux-x86_64 - ghc-8.10
git rev f1ce770834bf7150ca29cb647065c9e62d39be1a
```
**PROBLEM STATEMENT:** 

With the new stake address registration certificate type7 in conway era, we can now set a deposit amount. That exact deposit amount should also be used to generate the de-registration/retire certificate. For that we would need a method to query the registered amount. Best place imo would be within the stake-address-info query.

Currently we get something like:
```json
[
  {
    "address": "stake_test1urqntq4wexjylnrdnp97qq79qkxxvrsa9lcnwr7ckjd6w0cr04y4p",
    "delegation": "pool1rkfs9glmfva3jd0q9vnlqvuhnrflpzj4l07u6sayfx5k7d788us",
    "rewardAccountBalance": 297059772555
  }
]
```

So adding a new entry `deposited` would be a good solution imo to query the correct value before generating the de-registration/retire certificate.

P.S.: Same goes for the dRep information query. That should also list the deposited amount if possible.

carlhammann:
# Changelog

```yaml
- description: |
    `cardano-cli query stake-address-info` now returns the deposit for the currently delegated staking credential (in addition to what it returned before).
  type:
```

# Context

This PR closes #353.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -902,12 +902,19 @@ runQueryStakeAddressInfoCmd
 
         requireNotByronEraInByronMode eraInMode
 
-        result <- lift (queryStakeAddresses eInMode sbe stakeAddr networkId)
+        (stakeRewardAccountBalances, stakePools) <- lift (queryStakeAddresses eInMode sbe stakeAddr networkId)
           & onLeft (left . QueryCmdUnsupportedNtcVersion)
           & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
 
-        pure $ do
-          writeStakeAddressInfo mOutFile $ DelegationsAndRewards result
+        stakeDelegDeposits <- lift (queryStakeDelegDeposits eInMode sbe stakeAddr)
+          & onLeft (left . QueryCmdUnsupportedNtcVersion)
+          & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
+
+        return $ do
+          writeStakeAddressInfo
+            mOutFile
+            (DelegationsAndRewards (stakeRewardAccountBalances, stakePools))
+            (Map.mapKeys (makeStakeAddress networkId) stakeDelegDeposits)
     )
     & onLeft (left . QueryCmdAcquireFailure)
     & onLeft left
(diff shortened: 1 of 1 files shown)


Jimbo4350 (approved):
I like the improvement but can you inform QA about this change. They may depend on the output.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -902,12 +902,19 @@ runQueryStakeAddressInfoCmd
...
           & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
 
-        pure $ do
-          writeStakeAddressInfo mOutFile $ DelegationsAndRewards result
+        stakeDelegDeposits <- lift (queryStakeDelegDeposits eInMode sbe stakeAddr)
+          & onLeft (left . QueryCmdUnsupportedNtcVersion)
+          & onLeft (left . QueryCmdLocalStateQueryError . EraMismatchError)
+
+        return $ do
+          writeStakeAddressInfo
```
newhoggy, line 914: Is it worth updating the function name to say `writeStakeAddressInfoAndBalances`?

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -917,13 +924,42 @@ runQueryStakeAddressInfoCmd
...
-        $ LBS.writeFile fpath (encodePretty delegsAndRewards)
+writeStakeAddressInfo
+  mOutFile
+  (DelegationsAndRewards (stakeAccountBalances, stakePools))
+  stakeDelegDeposits =
+    case mOutFile of
+      Nothing -> liftIO $ LBS.putStrLn (encodePretty jsonInfo)
+      Just (File fpath) ->
+        handleIOExceptT (QueryCmdWriteFileError . FileIOError fpath)
+          $ LBS.writeFile fpath (encodePretty jsonInfo)
```
newhoggy: You should be able to use this function here? https://github.com/input-output-hk/cardano-api/blob/newhoggy/new-shelleyToAlonzoEraToShelleyToBabbageEra-function/cardano-api/internal/Cardano/Api/IO.hs/#L115-L123

newhoggy (approved):

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
carlhammann: Not sure; the name you propose wouldn't even have been completely descriptive before this change, but now we write balances, stake pools, and deposits... In short, all information associated with a stake address that we can query.

PR #385, opened 2023-10-18
Integrate latest ledger, consensus and api for 8.6.0

teodanciu:
# Changelog

```yaml
- description: |
    Updated cardano-ledger, ouroboros-consensus and cardano-api packages 
    Replaced  queryCommitteState with new queryCommitteeMembersState
    Added anchor to committee cold key resignation certificate command

  type:
```

# Context

Additional context for the PR goes here.

If the PR fixes a particular issue please provide a
[link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=)
to the issue.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
@@ -13,6 +13,8 @@ module Cardano.CLI.EraBased.Commands.Governance.Committee
   ) where
 
 import           Cardano.Api
+import qualified Cardano.Api.Ledger as Ledger
+import           Cardano.Api.Shelley
 
 import           Cardano.CLI.Types.Key
 import           Cardano.CLI.Types.Key.VerificationKey
@@ -59,6 +61,7 @@ data GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs era =
   GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs
     { eon               :: !(ConwayEraOnwards era)
     , vkeyColdKeySource :: !(VerificationKeyOrHashOrFile CommitteeColdKey)
+    , anchor            :: !(Maybe (Ledger.Anchor (Ledger.EraCrypto (ShelleyLedgerEra era))))
     , outFile           :: !(File () Out)
     } deriving Show
 
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
@@ -2560,7 +2560,6 @@ pProtocolParametersUpdate =
     <*> optional pPoolInfluence
     <*> optional pMonetaryExpansion
     <*> optional pTreasuryExpansion
-    <*> pure Nothing
     <*> pure mempty
     <*> optional pExecutionUnitPrices
     <*> optional pMaxTxExecutionUnits
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
@@ -2077,30 +2077,62 @@ pInvalidBefore = fmap SlotNo $ asum
     ]
   ]
 
-pInvalidHereafter :: Parser SlotNo
-pInvalidHereafter =
+pLegacyInvalidHereafter :: Parser SlotNo
+pLegacyInvalidHereafter =
   fmap SlotNo $ asum
-  [ Opt.option (bounded "SLOT") $ mconcat
-    [ Opt.long "invalid-hereafter"
-    , Opt.metavar "SLOT"
-    , Opt.help "Time that transaction is valid until (in slots)."
-    ]
-  , Opt.option (bounded "SLOT") $ mconcat
-    [ Opt.long "upper-bound"
-    , Opt.metavar "SLOT"
-    , Opt.help $ mconcat
-      [ "Time that transaction is valid until (in slots) "
-      , "(deprecated; use --invalid-hereafter instead)."
+    [ Opt.option (bounded "SLOT") $ mconcat
+      [ Opt.long "invalid-hereafter"
+      , Opt.metavar "SLOT"
+      , Opt.help "Time that transaction is valid until (in slots)."
       ]
-    , Opt.internal
+    , Opt.option (bounded "SLOT") $ mconcat
+      [ Opt.long "upper-bound"
+      , Opt.metavar "SLOT"
+      , Opt.help $ mconcat
+        [ "Time that transaction is valid until (in slots) "
+        , "(deprecated; use --invalid-hereafter instead)."
+        ]
+      , Opt.internal
+      ]
+    , Opt.option (bounded "SLOT") $ mconcat
+      [ Opt.long "ttl"
+      , Opt.metavar "SLOT"
+      , Opt.help "Time to live (in slots) (deprecated; use --invalid-hereafter instead)."
+      , Opt.internal
+      ]
     ]
-  , Opt.option (bounded "SLOT") $ mconcat
-    [ Opt.long "ttl"
-    , Opt.metavar "SLOT"
-    , Opt.help "Time to live (in slots) (deprecated; use --invalid-hereafter instead)."
-    , Opt.internal
-    ]
-  ]
+
+pInvalidHereafter :: ()
+  => CardanoEra era
+  -> Parser (TxValidityUpperBound era)
+pInvalidHereafter =
+  caseByronOrShelleyBasedEra
+    (pure . TxValidityNoUpperBound)
+    (\eon ->
+      fmap (TxValidityUpperBound eon) $ asum
+        [ fmap (Just . SlotNo) $ Opt.option (bounded "SLOT") $ mconcat
+          [ Opt.long "invalid-hereafter"
+          , Opt.metavar "SLOT"
+          , Opt.help "Time that transaction is valid until (in slots)."
+          ]
+        , fmap (Just . SlotNo) $ Opt.option (bounded "SLOT") $ mconcat
+          [ Opt.long "upper-bound"
+          , Opt.metavar "SLOT"
+          , Opt.help $ mconcat
+            [ "Time that transaction is valid until (in slots) "
+            , "(deprecated; use --invalid-hereafter instead)."
+            ]
+          , Opt.internal
+          ]
+        , fmap (Just . SlotNo) $ Opt.option (bounded "SLOT") $ mconcat
+          [ Opt.long "ttl"
+          , Opt.metavar "SLOT"
+          , Opt.help "Time to live (in slots) (deprecated; use --invalid-hereafter instead)."
+          , Opt.internal
+          ]
+        , pure Nothing
+        ]
+    )
 
 pTxFee :: Parser Lovelace
 pTxFee =
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
@@ -15,6 +15,7 @@ import           Cardano.CLI.EraBased.Commands.Query
 import           Cardano.CLI.EraBased.Options.Common
 import           Cardano.CLI.Types.Common
 
+import qualified Data.Text as Text (toLower, unpack)
 import           Data.Foldable
 import           Options.Applicative hiding (help, str)
 import qualified Options.Applicative as Opt
@@ -351,8 +352,33 @@ pQueryGetCommitteeStateCmd era envCli = do
   w <- forEraMaybeEon era
   pure
     $ subParser "committee-state"
-    $ Opt.info (QueryCommitteeStateCmd <$> pQueryNoArgCmdArgs w envCli)
+    $ Opt.info (QueryCommitteeMembersStateCmd <$> pQueryCommitteeMembersStateArgs w)
     $ Opt.progDesc "Get the committee state"
+  where
+    pQueryCommitteeMembersStateArgs :: ConwayEraOnwards era -> Parser (QueryCommitteeMembersStateCmdArgs era)
+    pQueryCommitteeMembersStateArgs w = QueryCommitteeMembersStateCmdArgs w
+      <$> pSocketPath envCli
+      <*> pConsensusModeParams
+      <*> pNetworkId envCli
+      <*> many pCommitteeColdVerificationKeyOrHashOrFile
+      <*> many pCommitteeHotKeyOrHashOrFile
+      <*> many pMemberStatus
+      <*> optional pOutputFile
+    pMemberStatus :: Parser MemberStatus
+    pMemberStatus =
+        Opt.option readerMemberStatus $ mconcat
+          [ Opt.long "memberstatus"
+          , Opt.metavar "MEMBER_STATUS"
+          , Opt.help "Member status filter: active/expired/unrecognized"
+          ]
+    readerMemberStatus :: ReadM MemberStatus
+    readerMemberStatus = do
+      v <- Opt.str
+      case Text.toLower v of
+        "active" -> pure Active
+        "expired" -> pure Expired
+        "unrecognized" -> pure Unrecognized
+        _ -> fail ("Unrecognized status: " <> Text.unpack v)
 
 pQueryNoArgCmdArgs :: ()
   => ConwayEraOnwards era
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
@@ -8,6 +8,7 @@ module Cardano.CLI.EraBased.Options.Query
   ) where
 
 import           Cardano.Api hiding (QueryInShelleyBasedEra (..))
+import qualified Cardano.Api as MemberStatus (MemberStatus (..))
 import           Cardano.Api.Shelley hiding (QueryInShelleyBasedEra (..))
 
 import           Cardano.CLI.Environment (EnvCli (..))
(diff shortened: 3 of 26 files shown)


s
```diff
@@ -0,0 +1,10 @@
+#!/run/current-system/sw/bin/bash
+for x in $(git ls-tree --full-tree --name-only -r HEAD cardano-cli); do
+    if [ "${x##*.}" == "hs" ]; then
+      if grep -qE '^#' $x; then
+        echo "$x contains CPP.  Skipping."
+      else
+        stylish-haskell -i $x
+      fi
+    fi
+  done
```
newhoggy: Is this script meant to be included?

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1527,9 +1527,8 @@ runQueryCommitteeState
   eraInMode <- toEraInMode cEra cMode
     & hoistMaybe (QueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) (AnyCardanoEra cEra))
 
-  committeeState <- runQuery localNodeConnInfo $ queryCommitteeState eraInMode sbe
-  writeOutput mOutFile $
-    Map.assocs $ committeeState ^. Ledger.csCommitteeCredsL
+  committeeState <- runQuery localNodeConnInfo $ queryCommitteeMembersState eraInMode sbe Set.empty Set.empty Set.empty
+  liftIO . putStr $ show committeeState
```
newhoggy: What's in the committee state?

Would it be useful to be able to write it to a file?

Also wondering if it's worth having a JSON encoding of it for consistency?

s
teodanciu: Sorry, I pushed it by mistake!  will remove

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
teodanciu: You're right,  no reason to not write it in the output file as it was before.  I will change it.

The result of the query  ([CommitteeMembersState](https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query/CommitteeMembersState.hs#L141)) is a more sophisticated version of CommitteeState, containing more info about the members, such as status, expiry, hot credential status.  Basically it's putting together information about committee members, from both sources: Committee (which is the current committee cold creds)  and CommitteState  (which is the mapping of cold to hot credentials of committee members).

> Also wondering if it's worth having a JSON encoding of it for consistency?

The type representing the result of this   query (CommitteeMembersState) [does have a JSON encoding defined in ledger. ](https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query/CommitteeMembersState.hs#L165)
So I suppose your suggestion is to use it, but could you explain where? Is there an example of this?

teodanciu:
Force-pushed to remove script added by mistake.

s
teodanciu: Removed.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
teodanciu: Oh I guess you mean to write the  json output in the file, rather than the result of Show. Seems like a good idea to me

teodanciu:
Force pushed to integrate with cardano-api-27.0

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
```diff
@@ -351,8 +351,25 @@ pQueryGetCommitteeStateCmd era envCli = do
...
+      <$> pSocketPath envCli
+      <*> pConsensusModeParams
+      <*> pNetworkId envCli
+      <*> many pCommitteeColdVerificationKeyOrHashOrFile
+      <*> many pCommitteeHotKeyOrHashOrFile
+      <*> many pMemberStatus
+      <*> optional pOutputFile
+    pMemberStatus :: Parser MemberStatus
+    pMemberStatus =
+        Opt.option Opt.auto $ mconcat
```
teodanciu: Actually this will change, because we don't want the `Read` instance in `MemberStatus`.  Will  change it shortly.

teodanciu:
Two golden tests are failing, because of the new options in the query - trying to fix them

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
```diff
@@ -351,8 +352,33 @@ pQueryGetCommitteeStateCmd era envCli = do
...
+      <*> optional pOutputFile
+    pMemberStatus :: Parser MemberStatus
+    pMemberStatus =
+        Opt.option readerMemberStatus $ mconcat
+          [ Opt.long "memberstatus"
+          , Opt.metavar "MEMBER_STATUS"
+          , Opt.help "Member status filter: active/expired/unrecognized"
+          ]
+    readerMemberStatus :: ReadM MemberStatus
+    readerMemberStatus = do
```
teodanciu: Is there a standard way to parse an enum without `Read`? Couldn't find anything similar in the codebase

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
```diff
@@ -59,6 +61,7 @@ data GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs era =
   GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs
     { eon               :: !(ConwayEraOnwards era)
     , vkeyColdKeySource :: !(VerificationKeyOrHashOrFile CommitteeColdKey)
+    , anchor            :: !(Maybe (Ledger.Anchor (Ledger.EraCrypto (ShelleyLedgerEra era))))
```
Jimbo4350, line 64: Have a look at how the governance action parsers handle anchors. E.g
```
data GovernanceActionCreateNoConfidenceCmdArgs era
  = GovernanceActionCreateNoConfidenceCmdArgs
      { eon                   :: !(ConwayEraOnwards era)
      , networkId             :: !Ledger.Network
      , deposit               :: !Lovelace
      , returnStakeAddress    :: !AnyStakeIdentifier
      , proposalUrl           :: !ProposalUrl
      , proposalHashSource    :: !ProposalHashSource
      , governanceActionId    :: !TxId
      , governanceActionIndex :: !Word32
      , outFile               :: !(File () Out)
      } deriving Show
```
In particular the `proposalUrl` and `proposalHashSource` fields.

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
Jimbo4350: What about making this a flag? I.e `many pMemberStatus` where `pMemberStatus` is something like:
```
pMemberStatus = asum [ Opt.flag' Ledger.Active .......]
```

cardano-cli/src/Cardano/CLI/Json/Friendly.hs
```diff
@@ -429,10 +429,12 @@ renderCertificate sbe = \case
                 [ "cold key hash" .= ck
                 , "hot key hash" .= hk
                 ]
-        Ledger.ResignCommitteeColdTxCert cred -> case cred of
+        Ledger.ResignCommitteeColdTxCert cred anchor -> case cred of
           Shelley.ScriptHashObj sh ->
             "Cold committee resignation" .= object
-              [ "script hash" .=  sh ]
+              [ "script hash" .=  sh
+              , "achor" .= anchor
```
Jimbo4350: spelling "anchor"

cardano-cli/src/Cardano/CLI/Types/Errors/QueryCmdError.hs
```diff
@@ -94,3 +96,7 @@ renderQueryCmdError = \case
   QueryCmdConvenienceError qce -> renderQueryConvenienceError qce
   QueryCmdDRepKeyError e ->
     "Error reading delegation representative key: " <> Text.pack (displayError e)
+  QueryCmdColdKeyError e ->
+    "Error reading cold key: " <> Text.pack (displayError e)
```
Jimbo4350: Copy paste error. We should also distinguish between cold and hot _comittee_ keys in the error message.

Jimbo4350 (commented):
This is looking good, a couple minor comments

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
teodanciu: @Jimbo4350  So you prefer to flatten these two fields , url and hash. 
But what to do about `ProposalUrl` and `ProposalHashSource` which  are not accurately describing this Anchor, no - since it's not related to proposal, but to resignation..  Create new types for them?

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
teodanciu: Yes! This is exactly what I was missing. So much better.
Made this change in a new commit.

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -2077,30 +2077,62 @@ pInvalidBefore = fmap SlotNo $ asum
...
-    [ Opt.long "ttl"
-    , Opt.metavar "SLOT"
-    , Opt.help "Time to live (in slots) (deprecated; use --invalid-hereafter instead)."
-    , Opt.internal
     ]
-  ]
+
+pInvalidHereafter :: ()
+  => CardanoEra era
+  -> Parser (TxValidityUpperBound era)
```
newhoggy, line 2107: In era-based commands we want to bring the era-sensitive types closer to the CLI.

This means we parse `TxValidityUpperBound era` instead of `Maybe SlotNo`.

This allows the parser to determine if the options are valid for the given era.

For example `pInvalidHereafter` will add these options only in a Shelley based eras and will be absent for the `Byron` era.

In the Byron era, `TxValidityNoUpperBound` always returned and there are no options for it.

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -2077,30 +2077,62 @@ pInvalidBefore = fmap SlotNo $ asum
     ]
   ]
 
-pInvalidHereafter :: Parser SlotNo
-pInvalidHereafter =
+pLegacyInvalidHereafter :: Parser SlotNo
```
newhoggy, line 2080: But we also have legacy commands that don't understand eras the same way.  In the legacy era, options are always there regardless of era and these need validation.  This is why these parse plain types, not era sensitive types.


This means we may receive values from the CLI parser that are invalid for our era and hence we need validation.

See https://github.com/input-output-hk/cardano-cli/pull/385/commits/85f657b566e835635aabbc78cc47a21e6a8a582b#r1374853935

cardano-cli/src/Cardano/CLI/Legacy/Run/Transaction.hs
```diff
@@ -146,10 +149,19 @@ runLegacyTransactionBuildRawCmd
...
       & firstExceptT TxCmdTxUpdateProposalValidationError
 
+  upperBound <-
+    caseByronOrShelleyBasedEra
+      (\w -> case mUpperBound of
+        Nothing -> pure $ TxValidityNoUpperBound w
+        Just _ -> left $ TxCmdTxValidityUpperBoundValidationError $ TxValidityUpperBoundNotSupported anyEra
+      )
+      (\w -> pure $ TxValidityUpperBound w mUpperBound)
+      era
```
newhoggy, line 158: For the legacy commands we require validation so we do it here.  If the era does not support a value, and a value was parsed by the CLI we raise an error.

cardano-cli/src/Cardano/CLI/Json/Friendly.hs
```diff
@@ -188,21 +188,11 @@ friendlyExtraKeyWits = \case
...
   TxExtraKeyWitnesses _supported paymentKeyHashes -> toJSON paymentKeyHashes
 
--- | Special case of validity range:
--- in Shelley, upper bound is TTL, and no lower bound
-pattern ShelleyTtl
-  :: Maybe SlotNo -> (TxValidityLowerBound era, TxValidityUpperBound era)
-pattern ShelleyTtl ttl <-
-  ( TxValidityNoLowerBound
-  , TxValidityUpperBound _ ttl
-  )
```
newhoggy, line 197: We don't do this anymore.  Allegra called the upper bound field "ttl" and had slightly different representation.  It represented no limit as `maxBound`.  Future eras represented no limit as `Nothing` instead.  We put a compatibility shim to make Allegra seem like it has the same presentation  - ie no limit means `Nothing` so we no longer have to do anything special here.

cardano-cli/test/cardano-cli-golden/files/golden/allegra/transaction-view.out
```diff
@@ -22,5 +22,6 @@ return collateral: null
 total collateral: null
 update proposal: null
 validity range:
-  time to live: 101
+  lower bound: null
+  upper bound: 101
```
newhoggy, line 26: As a result of [this](https://github.com/input-output-hk/cardano-cli/pull/385/commits/01e98c0fa081ddd94daa667a465f6eb4ccf0c639#r1374862298), we also show upper bound and never show time to live in `transaction view`.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -420,8 +420,6 @@ runTxBuildRaw era
       <- first TxCmdTxFeeValidationError $ validateTxFee era mFee
     validatedLowerBound
       <- first TxCmdTxValidityLowerBoundValidationError (validateTxValidityLowerBound era mLowerBound)
-    validatedUpperBound
-      <- first TxCmdTxValidityUpperBoundValidationError (validateTxValidityUpperBound era mUpperBound)
```
newhoggy: We no longer do validation here because by virtue of using era-sensitive types in the era-based CLI parser the values are already valid.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -532,7 +530,6 @@ runTxBuild
     <- hoistEither $ first TxCmdReturnCollateralValidationError $ validateTxReturnCollateral era mReturnCollateral
   dFee <- hoistEither $ first TxCmdTxFeeValidationError $ validateTxFee era dummyFee
   validatedLowerBound <- hoistEither (first TxCmdTxValidityLowerBoundValidationError (validateTxValidityLowerBound era mLowerBound))
-  validatedUpperBound <- hoistEither (first TxCmdTxValidityUpperBoundValidationError (validateTxValidityUpperBound era mUpperBound))
```
newhoggy: We no longer do validation here because by virtue of using era-sensitive types in the era-based CLI parser the values are already valid.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
teodanciu: Updated to write it as json.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
teodanciu: Does this mean I should remove this line `validateUpperBound <-... ` ?

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
teodanciu: Does this mean I should remove this line?  `validatedUpperBound <- ..` ?

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
Jimbo4350: So you can construct an `Anchor` as follows: https://github.com/input-output-hk/cardano-cli/blob/e86aa60f4481dcdc035d2173d61090bd284ad089/cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/Actions.hs#L98

teodanciu: Hm, I don't understand exactly what the suggestion is (probably because I'm not familiar with the codebase). 
Would you like me to represent Anchor  in  `GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs` as  two different fields, similar to the example you gave above, so as : 

```
   ,    resignationlUrl           :: !ProposalUrl
      , resignationHashSource    :: !ProposalHashSource
```

instead of the field that I have added: 
```
, anchor            :: !(Maybe (Ledger.Anchor (Ledger.EraCrypto (ShelleyLedgerEra era))))
```

? 

Did i understand correctly?

Jimbo4350: ~Yes exactly :+1:~ I see what you're saying now.

teodanciu: Both of them will have to be `Maybe` I suppose, since the field is optional.  I hope it won't be too messy, if say one is there and the other one not  :thinking:

Jimbo4350 (approved):
LGTM! Just get @CarlosLopezDeLara to sign off on the new parser names.

CarlosLopezDeLara (approved):

PR #392, opened 2023-10-19
Delete `AlonzEraOnly` and the ability to use datum-hash in the Alonzo era

newhoggy:
# Changelog

```yaml
- description: |
    Delete `AlonzEraOnly` and the ability to use datum-hash in the Alonzo era
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -729,18 +729,9 @@ toTxOutInAnyEra era (TxOutAnyEra addr' val' mDatumHash refScriptFp) = do
...
   val <- hoistEither $ toTxOutValueInAnyEra era val'
   (datum, refScript)
-    <- caseByronToMaryOrAlonzoEraOnwards
+    <- caseByronToAlonzoOrBabbageEraOnwards
          (const $ pure (TxOutDatumNone, ReferenceScriptNone))
-         (\w ->
-            caseAlonzoOnlyOrBabbageEraOnwards
-              (\wa ->
-                (,)
-                  <$> toTxAlonzoDatum (alonzoEraOnlyToAlonzoEraOnwards wa) mDatumHash
```
newhoggy: Previously we could use datum-hashes from Alonzo era onwards.  If we don't support that use case, the code simplifies.

PR #396, opened 2023-10-21
Remove `--update-proposal-file` from Conway onwards

newhoggy:
# Changelog

```yaml
- description: |
    Remove `--update-proposal-file` from Conway onwards
  type:
```

# Context

This option is not supported from Conway onwards to remove it from the CLI parser rather than accept it and then fail validation.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
@@ -3246,6 +3246,18 @@ pTxId l h =
 -- Helpers
 --------------------------------------------------------------------------------
 
+pFeatured :: ()
+  => Eon eon
+  => ToCardanoEra peon
+  => peon era
+  -> Parser a
+  -> Parser (Maybe (Featured eon era a))
+pFeatured peon p = do
+  let mw = forEraMaybeEon (toCardanoEra peon)
+  case mw of
+    Nothing -> pure Nothing
+    Just eon' -> Just . Featured eon' <$> p
+
 hiddenSubParser :: String -> ParserInfo a -> Parser a
 hiddenSubParser availableCommand pInfo =
   Opt.hsubparser $ Opt.command availableCommand pInfo <> Opt.metavar availableCommand <> Opt.hidden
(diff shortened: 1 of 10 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -3246,6 +3246,18 @@ pTxId l h =
 -- Helpers
 --------------------------------------------------------------------------------
 
+pFeatured :: ()
+  => Eon eon
+  => ToCardanoEra peon
+  => peon era
+  -> Parser a
+  -> Parser (Maybe (Featured eon era a))
```
carlhammann, line 3254: I think this is a good trick. With the context of this PR it is clear what this function accomplishes, but I feel a comment to the effect that this parser _doesn't show up_ in the command line if the era is not in the eon is warranted.

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -342,6 +335,16 @@ validateTxScriptValidity era (Just scriptValidity) = do
...
 
+-- TODO legacy. This can be deleted when legacy commands are removed.
+validateUpdateProposalFile
+  :: CardanoEra era
+  -> Maybe UpdateProposalFile
+  -> Either TxUpdateProposalValidationError (Maybe (Featured ShelleyToBabbageEra era (Maybe UpdateProposalFile)))
+validateUpdateProposalFile era = \case
+  Nothing -> pure Nothing
+  Just updateProposal -> do
+    supported <- conjureWitness era TxUpdateProposalNotSupported
```
carlhammann, line 346: Ah, there it is. I was wondering why `TxUpdateProposalNotSupported` wasn't deleted. :smile:

carlhammann (approved):
LGTM

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
newhoggy: I think documentation is good, but I didn't want to repeat myself by for example writing the same comment for every such parser we define.  Is there something else we can do?

carbolymer (approved):

CarlosLopezDeLara (approved):

Jimbo4350 (approved):

PR #435, opened 2023-11-05
Update to `cardano-8.31.0.0`

### LINKED ISSUE #388: remove options from create-protocol-parameters-update
Currently, `cardano-cli conway governance action create-protocol-parameters-update` gives the option of creating a proposal to change: 

```
--protocol-major-version  
--protocol-minor-version  
```

protocol-major is used to trigger a hardfork See [CDDL](https://github.com/input-output-hk/cardano-ledger/blob/b5753c7c3cc8102dae09ce157a260d091e00bfea/eras/conway/test-suite/cddl-files/conway.cddl#L102C58-L102C74) This kind of protocol parameter change is reserved to different type of governance action (Hardfork initiation), thus it makes no sense to have the option available under  create-protocol-parameters-update
 

Reference from CIP1694

`
> Separation of Hard Fork Initiation from Standard Protocol Parameter Changes
> In contrast to other protocol parameter updates, hard forks (or, more correctly, changes to the protocol's major version number) require much more attention. Indeed, while other protocol parameter changes can be performed without significant software changes, a hard fork assumes that a super-majority of the network has upgraded the Cardano node to support the new set of features that are introduced by the upgrade. This means that the timing of a hard fork event must be communicated well ahead of time to all Cardano users, and requires coordination between stake pool operators, wallet providers, DApp developers, and the node release team.
> 
> Hence, this proposal, unlike the Shelley scheme, promotes hard fork initiations as a standalone governance action, distinct from protocol parameter updates.

newhoggy:
# Changelog

```yaml
- description: |
    Update to `cardano-8.31.0.0`
  type:
```

# Context

Fixes https://github.com/input-output-hk/cardano-cli/issues/388

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-cli/test/cardano-cli-test/Test/Cli/MultiAssetParsing.hs
```diff
@@ -1,33 +1,36 @@
...
+    let value = fromLedgerValue ShelleyBasedEraConway ledgerValue
     tripping
       value
       renderValue
       (Parsec.parse parseValue "" . Text.unpack)
 
-hprop_roundtrip_Value_parse_renderPretty :: Property
-hprop_roundtrip_Value_parse_renderPretty =
+-- TODO enable these tests after switching completely to ledger types
+disable_hprop_roundtrip_Value_parse_renderPretty :: Property
```
Jimbo4350: Why can't we enable them now?

Jimbo4350 (approved):

cardano-cli/test/cardano-cli-test/Test/Cli/MultiAssetParsing.hs
newhoggy: The tests fail like when `Value` is completely empty and the generators are now capable of generating completely empty values:

```  roundtrip Value parse render:              FAIL
      ✗ Test.Cli.MultiAssetParsing.hprop_roundtrip_Value_parse_render failed at test/cardano-cli-test/Test/Cli/MultiAssetParsing.hs:23:5
        after 25 tests.
        shrink path: 25:
      
           ┏━━ test/cardano-cli-test/Test/Cli/MultiAssetParsing.hs ━━━
        18 ┃ hprop_roundtrip_Value_parse_render :: Property
        19 ┃ hprop_roundtrip_Value_parse_render =
        20 ┃   property $ do
        21 ┃     ledgerValue <- forAll $ genValueDefault MaryEraOnwardsConway
           ┃     │ MaryValue 0 (MultiAsset (fromList []))
        22 ┃     let value = fromLedgerValue ShelleyBasedEraConway ledgerValue
        23 ┃     tripping
           ┃     ^^^^^^^^
           ┃     │ ━━━ Original ━━━
           ┃     │ Right (valueFromList [])
           ┃     │ ━━━ Intermediate ━━━
           ┃     │ ""
           ┃     │ ━━━ Roundtrip ━━━
           ┃     │ Left (line 1, column 1):
           ┃     │ unexpected end of input
           ┃     │ expecting multi-asset value expression
        24 ┃       value
        25 ┃       renderValue
        26 ┃       (Parsec.parse parseValue "" . Text.unpack)
      
        This failure can be reproduced by running:
        > recheckAt (Seed 15168243033552485215 8658252526254707155) "25:" Test.Cli.MultiAssetParsing.hprop_roundtrip_Value_parse_render
```

newhoggy: I don't think we particularly care about empty `Value`, but for completeness, I think it good if the roundtrip worked for this edge case.  This can be fixed in `cardano-api`.

But since we're going to delete `Value` anyway, it probably isn't worth the effort.

In fact, I will delete these tests because they don't belong here:  They test nothing in `cardano-cli`.  See the import list for verification.

These tests should instead be moved to `cardano-api`.

PR #443, opened 2023-11-07
put n shelley/byron key witnesses arguments in correct order

brendanrbrown:
# Changelog

```yaml
- description: |
    Puts the `nShelleyKeyWitnesses`, `nByronKeyWitnesses` arguments to `runCalculateMinFeeCmd` in the correct order.
  type:
    - bugfix  
```

# Context

The `runCalculateMinFeeCmd` swaps the number of Byron and Shelley witnesses used to calculate fees, in its call to [estimateTransactionFee](https://github.com/input-output-hk/cardano-api/blob/95dbeb718106cce98452130018ba7dc4fc403b8e/cardano-api/internal/Cardano/Api/Fees.hs#L139) (see [here in the function's body](https://github.com/input-output-hk/cardano-api/blob/6d91127efc127c838689e011f67618d3cd2ffe4b/cardano-api/internal/Cardano/Api/Fees.hs#L156)) of `cardano-api-8.30.0.0`.

The witness numbers get multiplied by a factor that differs for the Byron and Shelley eras, and the Byron multiplier is larger than the Shelley number as defined [here](https://github.com/input-output-hk/cardano-api/blob/95dbeb718106cce98452130018ba7dc4fc403b8e/cardano-api/internal/Cardano/Api/Fees.hs#L171).

So the min fee printed by `runCalculateMinFeeCmd` will be larger than it should be if the number of Shelley witnesses exceeds the number Byron witnesses provided.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
@@ -969,7 +969,7 @@ runTransactionCalculateMinFeeCmd
                             (protocolParamTxFeePerByte pparams)
                             tx
                             nInputs nOutputs
-                            nByronKeyWitnesses nShelleyKeyWitnesses
+                            nShelleyKeyWitnesses nByronKeyWitnesses
 
       liftIO $ putStrLn $ (show fee :: String) <> " Lovelace"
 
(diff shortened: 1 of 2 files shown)


smelc:
Good catch! LGTM :+1:

smelc (approved):

brendanrbrown:
Great, thanks. FYI I don't have merge privileges, so one of you will have to push the button.

smelc:
FYI @Jimbo4350

PR #527, opened 2023-12-20
Provide more detailed error upon missing policy id

ghost:
When one gets this error while still providing one or more --mint-script-file, it can be confusing. This commit adds the list of "known" policy ids to the error message in order to help user troubleshoot their issue which might be related to an incorrectly computed hash or a wrong script.

# Changelog

```yaml
- description: |
    Provide more detailed error upon missing policy id
  type:
```

# Context

When a minting tx fails because the policy id is incorrect, one gets the following error message:

```
Command failed: transaction build  Error: The "--mint" flag specifies an asset with a policy Id, but no corresponding monetary policy script has been provided as a witness (via the "--mint-script-file" flag). The policy Id in question is: e18ad836532a69a93160efe11bcfac05b812a092ef3420042e700c10
```

When one gets this error while still providing one or more  `--mint-script-file`, it can be confusing. This commit adds the list of    "known" policy ids to the error message in order to help user    troubleshoot their issue which might be related to an incorrectly    computed hash or a wrong script.

The error message when a `--mint-script-file` argument is provided now says:

```
Command failed: transaction build  Error: The "--mint" flag specifies an asset with a policy Id, but no corresponding monetary policy script has been provided as a witness (via the "--mint-script-file" flag). The policy Id in question is: e18ad836532a69a93160efe11bcfac05b812a092ef3420042e700c10. Known policy Ids are: 994c90485cb725b3be04c6543326333261a0a043bb2a53e7fcbb65df
```

# How to trust this PR

This PR just modifies error messages displayed to users.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
+++ b/cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
@@ -49,6 +49,8 @@ data TxCmdError
   | TxCmdNotImplemented !Text
   | TxCmdWitnessEraMismatch !AnyCardanoEra !AnyCardanoEra !WitnessFile
   | TxCmdPolicyIdsMissing ![PolicyId] ![PolicyId]
+    -- The first list is the missing policy Ids, the second list is the
+    -- policy Ids that were provided in the transaction.
   | TxCmdPolicyIdsExcess  ![PolicyId]
   | TxCmdByronEra
   | TxCmdBalanceTxBody !TxBodyErrorAutoBalance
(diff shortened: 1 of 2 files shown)


cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
```diff
@@ -48,7 +48,7 @@ data TxCmdError
   | TxCmdTxBodyError !TxBodyError
   | TxCmdNotImplemented !Text
   | TxCmdWitnessEraMismatch !AnyCardanoEra !AnyCardanoEra !WitnessFile
-  | TxCmdPolicyIdsMissing ![PolicyId]
+  | TxCmdPolicyIdsMissing ![PolicyId] ![PolicyId]
```
smelc, line 51: Can you document what the two values stand for here? Since they have the same type, this declaration looks confusing without documentation.

smelc (approved):
LGTM modulo the need for an extra documentation line, so approving right away :champagne:

carbolymer (approved):

PR #537, opened 2024-01-02
Help hints at JSON formatting output with --out-file option

ghost:
# Changelog

Documents JSON-formatted output in CLI's help.

```yaml
- description: |
  type:
```

# Context

This is a very minor change to the documentation whose presence could have a saved me a few hours of head-banging. I wasn't aware (or forgot) one could get JSON-formatted output from the `query utxo` command and therefore coded a parser for UTxOs 🤦 

I am not sure where this information is readily available, but having it as part of the command-line help seems useful.

# How to trust this PR

🤔 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
@@ -1596,7 +1596,10 @@ pMaybeOutputFile =
   optional $ fmap File $ Opt.strOption $ mconcat
     [ Opt.long "out-file"
     , Opt.metavar "FILE"
-    , Opt.help "Optional output file. Default is to write to stdout."
+    , Opt.help "Optional output file. Default is to write to stdout. \
+               \If given, output will be formatted in JSON, otherwise\
+               \it will be formatted as human-readable text. Note that\
+               \using '/dev/stdout' outputs JSON."
     , Opt.completer (Opt.bashCompleter "file")
     ]
 


carbolymer (approved):
Thanks. You need to delete and regenerate golden file `test/cardano-cli-golden/files/golden/help/shelley_address_key-hash.cli` to make the test pass:
```bash
CREATE_GOLDEN_FILES=1 cabal test
```

ghost:
@carbolymer I am trying to fix the golden files but it's slightly annoying because I need to run the test, check the error, remove the file, and run again to get the next error. I wanted to refactor the test to use `conjoin` so that all golden checks are checked as a single property, in the hope of being able to get all the errors at once but this function is specific to QuickCheck. 
What would be the right way to do this with Hedgehog/Tasty?

newhoggy:
I propose to use `RECREATE_GOLDEN_FILES=1 cabal test`, the support for which is provided by this PR https://github.com/input-output-hk/hedgehog-extras/pull/54

This way you should be able to run the test once and manually spot check every golden file that was changed to check if the change was intended.

ghost:
🤔 Does `RECREATE_GOLDEN_FILES` recreates only the _failing_ golden files, or does it recreate all of them? Should I wait for the change to be merged, released and the library updated here to fix the PR?

carbolymer (changes requested):
I'm sorry, but I have to block this change. Our current serialization is in a bit messy state, so we're not guaranteeing that file will contain JSON and stdout will be human readable.

The description you're proposing would be misleading for some commands. An example: `cardano-cli query ledger-state`: https://github.com/IntersectMBO/cardano-cli/blob/17355f3e86dba264ca85389c491e561b2a07cafe/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs#L903 which behaves in an opposite way.

ghost:
@carbolymer I completely understand, no worries. Let me just close this and revisit at better times.

newhoggy:
> 🤔 Does RECREATE_GOLDEN_FILES recreates only the failing golden files, or does it recreate all of them?

It will recreate all of them, but the effect should be the same because passing golden tests will generate golden files that are identical to what is already there anyway.

ghost:
> It will recreate all of them, but the effect should be the same because passing golden tests will generate golden files that are identical to what is already there anyway.

Oh right, they are using the same seed 🤦

PR #541, opened 2024-01-03
Clarify stake-pool queries with an "all or some"-logic

carlhammann:
# Changelog

```yaml
- description: |
    Make queries that optionally filter their result by stake pool keys more explicit:
    - `cardano-cli * query-stake-snapshot`
    - `cardano-cli * query pool-params` (which is superseded by the next one, but still present)
    - `cardano-cli * query pool-state`
    - `cardano-cli conway query stake-snapshot`
  type:
```

# Context

This is a follow-up to #514, more concretely [this comment](https://github.com/IntersectMBO/cardano-cli/pull/514#discussion_r1423757149). I try to make it obvious from the flags and help texts what the logic of the command is. As an example, consider
```
Usage: cardano-cli conway query stake-snapshot      ...

                                                 ( --all-stake-pools
                                                 | (--stake-pool-id STAKE_POOL_ID)
                                                 )
                                                 ...

  Obtain the three stake snapshots for a pool, plus the total active stake
  (advanced command)

Available options:
  ...

  --all-stake-pools        Query for all stake pools
  --stake-pool-id STAKE_POOL_ID
                           Stake pool ID/verification key hash (either
                           Bech32-encoded or hex-encoded).

  ...
```

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


Jimbo4350 (commented):
I'm not sure I completely understand why the `Only` constructor was converted to `Some a [a]` but I think the semantics you are looking for can be expressed with `some`: https://github.com/IntersectMBO/cardano-cli/compare/main...jordan/clarify-all-or-some-flags.

Specifically: https://github.com/IntersectMBO/cardano-cli/commit/ee94ab5fce238672e7b2626fbc7addbaa93da2b0

carlhammann:
> I'm not sure I completely understand why the `Only` constructor was converted to `Some a [a]` but I think the semantics you are looking for can be expressed with `some`

@Jimbo4350 Not sure why I renamed `Only` to `Some` either... it's probably an artifact of an earlier attempt. ~~I'm aware of `some`, but optparse-applicative shows `some`, `many`, and `optional` in the same way, using `[...]`. I wanted to make really explicit that you'll have to provide *at least one* `--stake-pool-id`.~~

Edit: what I wrote above is untrue. `optional` and `many` are rendered the same, but that's not a problem here, since we only want to make visually obvious that `some` are needed. I'll follow your suggestion, then!

carlhammann:
@Jimbo4350 I changed to the simpler use of `some` in [04c23f6](https://github.com/IntersectMBO/cardano-cli/pull/541/commits/04c23f6d7c13ff5d939677c694f23e5def7aad9c), as you proposed. Thanks for clearing my confusion there!

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -635,7 +635,7 @@ runQueryPoolStateCmd
 
         let poolFilter = case allOrOnlyPoolIds of
               All -> Nothing
-              Only poolIds -> Just $ Set.fromList poolIds
+              Some poolId poolIds -> Just $ Set.fromList (poolId : poolIds)
```
Jimbo4350: This has the same issue we spoke about previously. We can rely on `some` vs `many` here.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
carlhammann: Are you sure you're looking at the most recent commit? I agreed with you on our previous discussion, and changed this is [4823900](https://github.com/IntersectMBO/cardano-cli/pull/541/commits/4823900f804a3b0c70831f09780768a736e4ec74).

Jimbo4350 (approved):

PR #557, opened 2024-01-08
Convenience flag to query stake distribution with `drep-state` query

### LINKED ISSUE #551: [FR] - include the stake-distribution within the drep-state query
To get the whole information about a drep keyHash, we have to make two queries:
1. query the `drep-state`
```json
[
    [
        {
            "keyHash": "2c0e9f599655bd3f7f313a3a61fe05fb9caac4cc3d1aa77740fc9579"
        },
        {
            "anchor": null,
            "deposit": 2000000,
            "expiry": 223
        }
    ]
]
```
2. query the `drep-stake-distribution`
```json
[
    [
        "drep-keyHash-2c0e9f599655bd3f7f313a3a61fe05fb9caac4cc3d1aa77740fc9579",
        1625348
    ]
]
```

Would it be possible to directly have an entry for the currently delegated amount within the `drep-state` output?

Like:
```json
[
    [
        {
            "keyHash": "2c0e9f599655bd3f7f313a3a61fe05fb9caac4cc3d1aa77740fc9579"
        },
        {
            "anchor": null,
            "deposit": 2000000,
            "expiry": 223
        },
        {
            "delegation": 1625348
        }
    ]
]
```
or similar?

carlhammann:
# Changelog

```yaml
- description: |
    Add `--include-stake` flag to obtain the stake in the `drep-state` query
  type:
```

# Context

This PR closes #551. 

I introduce an overlap between the `drep-state` and `drep-stake-distribution` queries. With the `--include-stake` flag, the former now subsumes the latter. AFAICT, the queries are exposed separately because [`drep-stake-distribution` is expensive](https://github.com/IntersectMBO/ouroboros-consensus/blob/404aaec8a4c8091594b4bc5a38e26f6ea9627f6b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs#L254). This makes obtaining all information associated with a DRep credential a two-command process. With the new flag, the expensive query is still guarded, but the CLI gets more user-friendly.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
@@ -335,6 +335,16 @@ pQueryDRepStateCmd era envCli = do
         <*> pConsensusModeParams
         <*> pNetworkId envCli
         <*> pAllOrOnlyDRepVerificationKeyOrHashOrFile
+        <*> Opt.flag WithStake NoStake (mconcat
+              [ Opt.long "include-stake"
+              , Opt.help $ mconcat
+                 [ "Also return the stake associated with each DRep. "
+                 , "The result is the same as with \"drep-stake-distribution\"; "
+                 , "this is a convenience option to obtain all information concerning a DRep at once. "
+                 , "This is a potentially expensive query."
+                 ]
+              ]
+            )
         <*> pTarget era
         <*> optional pOutputFile
 
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
@@ -1434,6 +1435,7 @@ runQueryDRepState
       , Cmd.consensusModeParams
       , Cmd.networkId
       , Cmd.drepKeys = drepKeys'
+      , Cmd.includeStake
       , Cmd.target
       , Cmd.mOutFile
       } = conwayEraOnwardsConstraints eon $ do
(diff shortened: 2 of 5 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
```diff
@@ -323,6 +323,16 @@ pQueryDRepStateCmd era envCli = do
         <*> pConsensusModeParams
         <*> pNetworkId envCli
         <*> many pDRepVerificationKeyOrHashOrFile
+        <*> Opt.flag True False (mconcat
```
Jimbo4350: Implement a sum type for this to avoid boolean blindness.

Jimbo4350 (commented):
LGTM but one comment

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
carlhammann: Good point, done!

gitmachtl:
looks like the current command defaults to `--include-stake`, i guess that should be false per default?

also, the generated output does not contain the keyhash anymore. looks like:

```json
[
    {
        "anchor": {
            "dataHash": "1111111111111111111111111111111111111111111111111111111111111111",
            "url": "https://bit.ly/3zCH2HL"
        },
        "deposit": 2000000,
        "expiry": 188,
        "stake": null
    },
    {
        "anchor": null,
        "deposit": 2000000,
        "expiry": 230,
        "stake": null
    },
    {
        "anchor": null,
        "deposit": 2000000,
        "expiry": 216,
        "stake": null
    },
    {
        "anchor": {
            "dataHash": "f8f21b94aef91b0d5e7ff330889d60f665529897b5c1ae4a1aa608b6249e20e6",
            "url": "https://github.com/Mborongoma/SANCH-GOV-TESTES/blob/main/bio.txt"
        },
        "deposit": 2000000,
        "expiry": 227,
        "stake": null
    },
    {
        "anchor": {
            "dataHash": "fd1b4e7fd844a8e2805bfda173f6461fd59812fcad4ce8a69b5ab427f7226cf7",
            "url": "https://my-ip.at/test/hwdrep.json"
        },
        "deposit": 2000000,
        "expiry": 232,
        "stake": null
    }
]
```

carlhammann:
> looks like the current command defaults to `--include-stake`, i guess that should be false per default?

Well spotted, @gitmachtl! The flag is actually off by default, in the sense that the _query_ isn't performed, but the _output_ includes the `"stake"` field always. That's confusing, so I changed that behaviour in [f11dd55](https://github.com/IntersectMBO/cardano-cli/pull/557/commits/f11dd55d72e311fcbabd88c89feba7d3e014c8e5): Now the field should only be present when `--include-stake` is used.

gitmachtl:
> > looks like the current command defaults to `--include-stake`, i guess that should be false per default?
> 
> Well spotted, @gitmachtl! The flag is actually off by default, in the sense that the _query_ isn't performed, but the _output_ includes the `"stake"` field always. That's confusing, so I changed that behaviour in [f11dd55](https://github.com/IntersectMBO/cardano-cli/pull/557/commits/f11dd55d72e311fcbabd88c89feba7d3e014c8e5): Now the field should only be present when `--include-stake` is used.

thx! and the keyHash is missing in the output too for some reason.

carlhammann:
> thx! and the keyHash is missing in the output too for some reason.

Fixed in [9bd912b](https://github.com/IntersectMBO/cardano-cli/pull/557/commits/9bd912b6d8de006caf993a40e103d53ed6e97bf1). Thanks for the reminder :bow:

gitmachtl:
hi, any plans to merge this into main soon? thx.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1419,24 +1420,40 @@ runQueryDRepState
...
   writeOutput mOutFile $
-    second drepStateToJson <$> Map.assocs drepState
+    drepStateToJson drepStakeDistribution <$> Map.assocs drepState
   where
-    drepStateToJson ds = A.object
-      [ "expiry" .= (ds ^. Ledger.drepExpiryL)
-      , "anchor" .= (ds ^. Ledger.drepAnchorL)
-      , "deposit" .= (ds ^. Ledger.drepDepositL)
-      ]
+    drepStateToJson stakeDistr (cred, ds) = (cred,) . A.object $
```
Jimbo4350, line 1460: This is fine for now but we should convert this into a `ToJSON` instance.

Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
smelc: Follow-up issue created: https://github.com/IntersectMBO/cardano-cli/issues/606

Issue #477, opened 2023-11-23
Missing tests for  online commands


carlhammann:
This is a pretty big one: Currently, AFAICT, all commands that need to interact with a running node have no tests. This group of commands contains
- all queries
- transaction building (with `build`, not `build-raw`)
- transaction submission

I'll split this issue up into smaller ones, but thought it worthwhile to have the big one as a reminder.

smelc:
There are a few such tests in cardano-node already, and we are in the process of adding more, as part of the Conway testing roadmap. So closing.

PR #569, opened 2024-01-16
Use (slightly hacked) cardano-ledger JSON output for protocol parameters

### LINKED ISSUE #554: [FR] - include governance parameters also in the protocol-parameters query
```console
$ cardano-cli version
cardano-cli 8.17.0.0 - linux-x86_64 - ghc-8.10
git rev a4a8119b59b1fbb9a69c79e1e6900e91292161e7
```

Is it possible to also include the current governance related parameters directly into the procotol-parameters query?

For example, currently we have to do a 
```bash
cardano-cli conway query protocol-parameters
```
action to get the general parameters for building a tx.

And when we do for example a DRep-ID registration we need to also do a
```bash
cardano-cli conway query gov-state | jq -r ".enactState.curPParams"
```
command to get mostly the same values (not all) and the additional ones to do correct transactions for the new conway stuff.

So it would be supernice if the current governance parameters would also be included in the protocol-parameters query to just have one query that shows all current parameters set.

carlhammann:
# Changelog

```yaml
- description: |
    Adapt the output of the `protocol-parameters` query to show all parameters in Conway. Includes a temporary hack around faulty ToJSON instance in `cardano-ledger`.
  type:
```

# Context

This closes #554. This issue was caused by a mismatch between the protocol parameters returned by `cardano-cli * query protocol-parameters` and `cardano-cli conway query gov-state`. Both commands retreive exactly the same parameters from the node, but they use different -- and differently wrong -- `ToJSON` instances:
- The `query * query protocol-parameters` command uses an [instance from cardano-api](https://github.com/IntersectMBO/cardano-api/blob/2b9596161f7c43d5f6928656595591d354c032e3/cardano-api/internal/Cardano/Api/ProtocolParameters.hs#L648), which doesn't include the Conway parameters.
- The other query uses the `ToJSON` instance from cardano-ledger, which includes all the Conway parameters. However, it is also faulty, in that it does not include the `protocolVersion` in Conway. 

The latter of these two instances is the one I think we should be using (so as to avoid code duplication in the cardano-api; I propose we try to get rid of that instance and the associated maintenance burden). The [commit that adds the `protocolVersion`](https://github.com/IntersectMBO/cardano-ledger/pull/3953/commits/df9ee19944099a75d75019b8b36e99b03db5b558) is already on merged on cardano-api, but not yet released (nor part of our dependencies).

That's why this PR contains a hack to add the `protocolVersion`, which will make the output correct now, and won't be harmful later on.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -56,6 +56,7 @@ import           Cardano.CLI.Types.Key
 import qualified Cardano.CLI.Types.Output as O
 import           Cardano.Crypto.Hash (hashToBytesAsHex)
 import qualified Cardano.Crypto.Hash.Blake2b as Blake2b
+import           Cardano.Ledger.Alonzo.Core (EraPParams (ppProtocolVersionL))
 import qualified Cardano.Ledger.BaseTypes as L
 import qualified Cardano.Ledger.Core as Core
 import qualified Cardano.Ledger.Credential as L
@@ -84,6 +85,7 @@ import           Control.Monad.Trans.Except.Extra
 import           Data.Aeson as Aeson
 import qualified Data.Aeson as A
 import           Data.Aeson.Encode.Pretty (encodePretty)
+import qualified Data.Aeson.KeyMap as Aeson
 import           Data.Bifunctor (Bifunctor (..))
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import           Data.Coerce (coerce)
@@ -184,23 +186,39 @@ runQueryProtocolParametersCmd
   AnyCardanoEra era <- firstExceptT QueryCmdAcquireFailure $ newExceptT $ determineEra localNodeConnInfo
   sbe <- forEraInEon @ShelleyBasedEra era (left QueryCmdByronEra) pure
   let qInMode = QueryInEra $ QueryInShelleyBasedEra sbe Api.QueryProtocolParameters
-  pp <- firstExceptT QueryCmdConvenienceError
-          . newExceptT $ executeQueryAnyMode localNodeConnInfo qInMode
-  writeProtocolParameters sbe mOutFile pp
+  pParams <- firstExceptT QueryCmdConvenienceError
+              . newExceptT $ executeQueryAnyMode localNodeConnInfo qInMode
+  writeProtocolParameters sbe mOutFile pParams
   where
-    -- TODO: Conway era - use ledger PParams JSON
     writeProtocolParameters
       :: ShelleyBasedEra era
       -> Maybe (File () Out)
       -> Ledger.PParams (ShelleyLedgerEra era)
       -> ExceptT QueryCmdError IO ()
     writeProtocolParameters sbe mOutFile' pparams =
-      let apiPParamsJSON = (encodePretty $ fromLedgerPParams sbe pparams)
-      in case mOutFile' of
-        Nothing -> liftIO $ LBS.putStrLn apiPParamsJSON
-        Just (File fpath) ->
-          handleIOExceptT (QueryCmdWriteFileError . FileIOError fpath) $
-            LBS.writeFile fpath apiPParamsJSON
+      firstExceptT QueryCmdWriteFileError . newExceptT $
+        writeLazyByteStringOutput mOutFile' (encodePretty $ toJSONWithPParams sbe pparams)
+      where
+        -- TODO: (written 2024-01-16) Currently, the ToJSON implementation for
+        -- Conway protocol parameters from cardano-ledger misses the
+        -- 'protocolVersion' field. The commit that fixes this is already on
+        -- main, but not part of a released version of cardano-ledger that we
+        -- can depend on. It's this one:
+        --
+        -- https://github.com/IntersectMBO/cardano-ledger/pull/3953/commits/df9ee19944099a75d75019b8b36e99b03db5b558
+        --
+        -- Until we have that in our dependencies, this workaround will ensure
+        -- we're printing the protocol version. It'll also do no harm after
+        -- that point, apart from heating the room a little.
+        toJSONWithPParams :: ShelleyBasedEra era -> Ledger.PParams (ShelleyLedgerEra era) -> Aeson.Value
+        toJSONWithPParams w pps = shelleyBasedEraConstraints w $
+          case toJSON pps of
+            Aeson.Object almostAllPairs ->
+              caseShelleyToBabbageOrConwayEraOnwards
+                (const . Aeson.Object $ almostAllPairs)
+                (const . Aeson.Object . Aeson.insert "protocolVersion" (toJSON $ pps ^. ppProtocolVersionL) $ almostAllPairs)
+                sbe
+            _ -> error "Expected the protocol parameters to be a 'KeyMap'. This is part of a temporary hack. Alert the cardano-cli team of this."
 
 -- | Calculate the percentage sync rendered as text.
 percentage


carlhammann:
@gitmachtl This is now the output of the command in Conway:
<details>

```json
{
    "a0": 0.0,
    "coinsPerUTxOByte": 4310,
    "collateralPercentage": 150,
    "committeeMaxTermLength": 60,
    "committeeMinSize": 0,
    "costmdls": {
        "PlutusV1": [
            197209,
            0,
            1,
            1,
            396231,
            621,
            0,
            1,
            150000,
            1000,
            0,
            1,
            150000,
            32,
            2477736,
            29175,
            4,
            29773,
            100,
            29773,
            100,
            29773,
            100,
            29773,
            100,
            29773,
            100,
            29773,
            100,
            100,
            100,
            29773,
            100,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            1000,
            0,
            1,
            150000,
            32,
            150000,
            1000,
            0,
            8,
            148000,
            425507,
            118,
            0,
            1,
            1,
            150000,
            1000,
            0,
            8,
            150000,
            112536,
            247,
            1,
            150000,
            10000,
            1,
            136542,
            1326,
            1,
            1000,
            150000,
            1000,
            1,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            1,
            1,
            150000,
            1,
            150000,
            4,
            103599,
            248,
            1,
            103599,
            248,
            1,
            145276,
            1366,
            1,
            179690,
            497,
            1,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            148000,
            425507,
            118,
            0,
            1,
            1,
            61516,
            11218,
            0,
            1,
            150000,
            32,
            148000,
            425507,
            118,
            0,
            1,
            1,
            148000,
            425507,
            118,
            0,
            1,
            1,
            2477736,
            29175,
            4,
            0,
            82363,
            4,
            150000,
            5000,
            0,
            1,
            150000,
            32,
            197209,
            0,
            1,
            1,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            150000,
            32,
            3345831,
            1,
            1
        ],
        "PlutusV2": [
            205665,
            812,
            1,
            1,
            1000,
            571,
            0,
            1,
            1000,
            24177,
            4,
            1,
            1000,
            32,
            117366,
            10475,
            4,
            23000,
            100,
            23000,
            100,
            23000,
            100,
            23000,
            100,
            23000,
            100,
            23000,
            100,
            100,
            100,
            23000,
            100,
            19537,
            32,
            175354,
            32,
            46417,
            4,
            221973,
            511,
            0,
            1,
            89141,
            32,
            497525,
            14068,
            4,
            2,
            196500,
            453240,
            220,
            0,
            1,
            1,
            1000,
            28662,
            4,
            2,
            245000,
            216773,
            62,
            1,
            1060367,
            12586,
            1,
            208512,
            421,
            1,
            187000,
            1000,
            52998,
            1,
            80436,
            32,
            43249,
            32,
            1000,
            32,
            80556,
            1,
            57667,
            4,
            1000,
            10,
            197145,
            156,
            1,
            197145,
            156,
            1,
            204924,
            473,
            1,
            208896,
            511,
            1,
            52467,
            32,
            64832,
            32,
            65493,
            32,
            22558,
            32,
            16563,
            32,
            76511,
            32,
            196500,
            453240,
            220,
            0,
            1,
            1,
            69522,
            11687,
            0,
            1,
            60091,
            32,
            196500,
            453240,
            220,
            0,
            1,
            1,
            196500,
            453240,
            220,
            0,
            1,
            1,
            1159724,
            392670,
            0,
            2,
            806990,
            30482,
            4,
            1927926,
            82523,
            4,
            265318,
            0,
            4,
            0,
            85931,
            32,
            205665,
            812,
            1,
            1,
            41182,
            32,
            212342,
            32,
            31220,
            32,
            32696,
            32,
            43357,
            32,
            32247,
            32,
            38314,
            32,
            35892428,
            10,
            9462713,
            1021,
            10,
            38887044,
            32947,
            10
        ]
    },
    "dRepActivity": 0,
    "dRepDeposit": 0,
    "dRepVotingThresholds": {
        "dvtCommitteeNoConfidence": 0.51,
        "dvtCommitteeNormal": 0.51,
        "dvtHardForkInitiation": 0.51,
        "dvtMotionNoConfidence": 0.51,
        "dvtPPEconomicGroup": 0.51,
        "dvtPPGovGroup": 0.51,
        "dvtPPNetworkGroup": 0.51,
        "dvtPPTechnicalGroup": 0.51,
        "dvtTreasuryWithdrawal": 0.51,
        "dvtUpdateToConstitution": 0.51
    },
    "eMax": 18,
    "govActionDeposit": 0,
    "govActionLifetime": 14,
    "keyDeposit": 0,
    "maxBlockBodySize": 65536,
    "maxBlockExUnits": {
        "exUnitsMem": 80000000,
        "exUnitsSteps": 40000000000
    },
    "maxBlockHeaderSize": 1100,
    "maxCollateralInputs": 3,
    "maxTxExUnits": {
        "exUnitsMem": 16000000,
        "exUnitsSteps": 10000000000
    },
    "maxTxSize": 16384,
    "maxValSize": 5000,
    "minFeeA": 44,
    "minFeeB": 155381,
    "minPoolCost": 0,
    "nOpt": 100,
    "poolDeposit": 0,
    "poolVotingThresholds": {
        "pvtCommitteeNoConfidence": 0.51,
        "pvtCommitteeNormal": 0.51,
        "pvtHardForkInitiation": 0.51,
        "pvtMotionNoConfidence": 0.51
    },
    "prices": {
        "prMem": 5.77e-2,
        "prSteps": 7.21e-5
    },
    "protocolVersion": {
        "major": 7,
        "minor": 0
    },
    "rho": 0.1,
    "tau": 0.1
}
```
</details>
Is this what you were looking for?

Jimbo4350 (approved):

gitmachtl:
@carlhammann thx!

ok, so the names for the "normal" parameters changed too. ok, should be an easy thing to fix in the 3rd party tools. can the cli now directly work with this protocol-parameter output when its provided as the protocol-parameters-file back to the cli for calculations?

as an example, before we had `txFeeFixed` thats now `minFeeB`. is the plan to bring back the "old" names or are we going with the new ones from now on?

btw, is this a mockup of the output, because conway era and protocol major-version 7?

gitmachtl:
@carlhammann when using the protocol parameters for a transaction/fee calculation. cli is complaining:

`key "extraPraosEntropy" not found`
`key "txFeeFixed" not found`
etc... 

so the input parser is currently not in line with the protocol-parameters output

ch1bo:
Good catch @gitmachtl!

@carlhammann @Jimbo4350 This is a breaking change on the `cardano-cli query protocol-parameters`, which is relied on by many users (including me via https://github.com/input-output-hk/hydra)!

If you deliberately want to break that interface, I do expect
* a CHANGELOG entry explaining this
* it to roundtrip with other places where the `cardano-cli` accepts protocol parameters as JSON

Both are not done, which suggests that this breakage was maybe not intended?

It's also surprising that parameters that are the same in conway would be serialized differently than before in babbage. i.e. The `cardano-cli` would change the output format on the hard-fork!

If you don't do it already (and ignored when merging this) I can **strongly** recommend to test this interface against accidental regressions!

As it happens, we recently had inconsistencies in our JSON schemas here as well and even started testing the `cardano-cli` protocol parameter output [here](https://github.com/input-output-hk/hydra/blob/7a19916b7c14e64d1b7ce0cac146101667323546/hydra-cluster/test/Test/Hydra/Cluster/CardanoCliSpec.hs#L50). The JSON schema that hydra expects from `cardano-cli` and also accepts on the command-line and provides to our users on our api is [this one](https://github.com/input-output-hk/hydra/blob/7a19916b7c14e64d1b7ce0cac146101667323546/hydra-node/json-schemas/api.yaml#L1059C24-L1059C24).

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -184,23 +186,39 @@ runQueryProtocolParametersCmd
...
+  writeProtocolParameters sbe mOutFile pParams
   where
-    -- TODO: Conway era - use ledger PParams JSON
     writeProtocolParameters
       :: ShelleyBasedEra era
       -> Maybe (File () Out)
       -> Ledger.PParams (ShelleyLedgerEra era)
       -> ExceptT QueryCmdError IO ()
     writeProtocolParameters sbe mOutFile' pparams =
-      let apiPParamsJSON = (encodePretty $ fromLedgerPParams sbe pparams)
```
ch1bo, line 198: If we don't want to break this interface, I think we should keep this `fromLedgerPParams` and do the patching on the cardano-api encoding of `ProtocolParameters`.

carlhammann:
@ch1bo @gitmachtl Thanks for the heads-up! It seems I underestimated the consequences of this change: the move towards using the cardano-ledger JSON encoding breaks
- round-tripping within the cardano-cli itself (I could have expected that, but only thought of it when @gitmachtl found the first instance...), and
- expectations of downsream users.

So, it seems we're faced with the two options I mentioned in the PR desciption:
- extend the cardano-api encoding. This would make it so that this is not a breaking change.
- use the cardano-ledger encoding. This would (eventually) mean less maintenance and less potential confusion, when and if we remove the cardano-api encoding.

At the very least, the post-merge comments show that this is not a change that we can take lightly, nor one that we can (thoughtlessly) do in small increments. It's above my pay grade to decide in which direction we proceed here. What do you think @Jimbo4350?

carlhammann:
> If you don't do it already (and ignored when merging this) I can **strongly** recommend to test this interface against accidental regressions!

That's a very valid point, @ch1bo, and a general problem with the test coverage of cardano-cli at the moment: we currently have no tests for "online" commands like queries. I opened an [issue documenting this](https://github.com/IntersectMBO/cardano-cli/issues/477) quite some time ago. The hope is to use [cardano-testnet](https://github.com/IntersectMBO/cardano-node/tree/master/cardano-testnet) for this purpose, and much of the team's work currently goes towards this.

carlhammann:
> btw, is this a mockup of the output, because conway era and protocol major-version 7?

@gitmachtl No, I produced it with a Conway test net, which I started using the scripts [here](https://github.com/IntersectMBO/cardano-node/tree/master/scripts/babbage) (which, confusingly, are still called "babbage"...)

gitmachtl:
> > btw, is this a mockup of the output, because conway era and protocol major-version 7?
> 
> @gitmachtl No, I produced it with a Conway test net, which I started using the scripts [here](https://github.com/IntersectMBO/cardano-node/tree/master/scripts/babbage) (which, confusingly, are still called "babbage"...)

ok, thats strange. i tested in on "sanchonet" and the output showed the expected protocolMajor=9

carlhammann:
Summarising the discussion on Slack with @Jimbo4350, @ch1bo, and @lehins:

It seems the most attractive way forward a is third option, which I didn't think of above: 
1. leave the cardano-api encoding alone for now,
2. change the  cardano-ledger encoding to match the current output, and
3. (at some point in the future) get rid of the cardano-api encoding.

This will ensure backwards compatibility __and__ reduce the amount of code. To make sure this works, there is
0. document the current state with golden tests.

How we do this zeroth step, in view of the problem I described [here](https://github.com/IntersectMBO/cardano-cli/pull/569#issuecomment-1895671910), is another question.

Agreed?

PR #617, opened 2024-02-22
query stake-pools: add --output-[json,text] flag to control format of the output

smelc:
# Changelog

```yaml
- description: |
    query stake-pools, add --output-[json,text] flag to control format of the output
  type:
```

# Context

* Part of fixing https://github.com/IntersectMBO/cardano-cli/issues/566
* Part of this stream of PRs: https://github.com/IntersectMBO/cardano-cli/pull/523, https://github.com/IntersectMBO/cardano-cli/pull/611
* Spin-off of https://github.com/IntersectMBO/cardano-node/commits/smelc/test-cli-queries/

# How to trust this PR

1. See is the output of `query stake-pools` before this PR (so no using the new flags): https://github.com/IntersectMBO/cardano-node/blob/smelc/test-cardano-cli-query-stake-pools-output-new/log.txt
1. See the output of `query stake-pools` after this PR, when not using the new flags, so clunky defaults: https://github.com/IntersectMBO/cardano-node/blob/smelc/test-cardano-cli-query-stake-pools-output-new/log_new.txt. Note it's the same as item 1.
2. See the output of `query stake-pools` after this PR, when using all combinations of new flags: https://github.com/IntersectMBO/cardano-node/blob/smelc/test-cardano-cli-query-stake-pools-output-new/log_all_new_cases.txt Notice the behaviors are the expected ones for all combinations.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -1126,6 +1126,7 @@ runQueryStakePoolsCmd
     , Cmd.consensusModeParams
     , Cmd.networkId
     , Cmd.target
+    , Cmd.format
     , Cmd.mOutFile
     } = do
   let localNodeConnInfo = LocalNodeConnectInfo consensusModeParams networkId nodeSocketPath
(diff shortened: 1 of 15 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1614,3 +1608,16 @@ requireEon minEra era =
...
+-- and that used to have the following default: --out-file implies JSON,
+-- output to stdout implied text.
+newOutputFormat :: Maybe QueryOutputFormat -> Maybe a -> QueryOutputFormat
+newOutputFormat format mOutFile =
+  case (format, mOutFile) of
+    (Just f, _) -> f -- Take flag from CLI if specified
+    (Nothing, Nothing) -> QueryOutputFormatText -- No CLI flag, writing to stdout: write text
+    (Nothing, Just _) -> QueryOutputFormatJson -- No CLI flag, writing to a file: write JSON
+
+textToLazyBytestring :: Text -> LBS.ByteString
```
Jimbo4350: What about: https://hackage.haskell.org/package/text-2.1.1/docs/Data-Text-Lazy-Encoding.html#v:encodeUtf8

Jimbo4350 (approved):
LGTM! One comment

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
smelc: This doesn't work, because my function converts _strict_ Text to a lazy bytestring.

So I changed the function's name to `strictTextToLazyBytestring` to make that more apparent. It was misleading :+1:

PR #642, opened 2024-03-11
Use evaluateTransactionFee in calculate-min-fee

newhoggy:
# Changelog

```yaml
- description: |
    Fix calculate-min-fee by switching to use Cardano.Api.evaluateTransactionFee
    instead of the deprecated Cardano.Api.estimateTransactionFee.
    This also deprecates the --mainnet, --testnet-magic, --tx-in-count,
    and --tx-out-count command args, which are no longer necessary.
    They can still be provided, but have no effect.
  type:
```

# Context

Original PR: https://github.com/IntersectMBO/cardano-cli/pull/534

`Cardano.Api.estimateTransactionFee` (currently used by Cardano CLI `calculate-min-fee`) tends to overestimate transaction fees. Furthermore, inspecting the code, `estimateTransactionFee` appears to only apply the old `a * x + b` fee formula to the transaction size (augmented to take into account the user-provided Shelley and Byron witness counts). This is not how fees should be calculated for later eras.

A deprecation note is currently attached to `Cardano.Api.estimateTransactionFee`, saying that `Cardano.Api.evaluateTransactionFee` should be used instead. Inspecting the code, evaluateTransactionFee calls the Cardano Ledger's era-aware fee calculation function `Cardano.Ledger.Shelley.API.Wallet.evaluateTransactionFee`, which in turn calls the `EraTx` class function `getMinFeeTx` that is instantiated for every Cardano Ledger era.

https://github.com/IntersectMBO/cardano-api/blob/1c10a287093ea1396c4c0ef0c92460f60bab690f/cardano-api/internal/Cardano/Api/Fees.hs#L190

Therefore, cardano-cli should be updated to use Cardano.Api.evaluateTransactionFee in the handler for calculate-min-fee.

https://twitter.com/amw7/status/1740008349905928326

# How to trust this PR

> Highlight important bits of the PR that will make the review faster. If there are commands the reviewer can run to observe the new behavior, describe them.

New golden test to calculate the min fee of a Babbage transaction body with Babbage protocol parameters, comparing against the known minimum fee (below which the transaction gets rejected when submitted to node).

https://github.com/IntersectMBO/cardano-cli/pull/534/files#diff-8d95bc46c5ef4bbab74b92c6b29635f1c3792d1e084d8c7668b0ce75bf9d2b52

# Checklist

- [x] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -2231,15 +2244,15 @@ pTxInCount =
   fmap TxInCount $ Opt.option Opt.auto $ mconcat
     [ Opt.long "tx-in-count"
     , Opt.metavar "NATURAL"
-    , Opt.help "The number of transaction inputs."
+    , Opt.help "DEPRECATED. This argument has no effect."
```
smelc, line 2247: Can you rename the enclosing function to `pTxInCountDeprecated`? Otherwise new callers could be tempted to use this function, where as they should not.

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -2231,15 +2244,15 @@ pTxInCount =
...
     ]
 
 pTxOutCount :: Parser TxOutCount
 pTxOutCount =
   fmap TxOutCount $ Opt.option Opt.auto $ mconcat
     [ Opt.long "tx-out-count"
     , Opt.metavar "NATURAL"
-    , Opt.help "The number of transaction outputs."
+    , Opt.help "DEPRECATED. This argument has no effect."
```
smelc, line 2255: Can you rename the enclosing function to `pTxOutCountDeprecated`? Otherwise new callers could be tempted to use this function, where as they should not.

carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
newhoggy: Done!

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
newhoggy: Done!

newhoggy:
CC @GeorgeFlerovsky @gitmachtl

GeorgeFlerovsky:
🎉

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -1009,46 +1012,94 @@ runTransactionCalculateMinFeeCmd :: ()
...
+                      (protocolParamTxFeePerByte pparams)
+                      nByronKeyWitnesses
+
+      let L.Coin fee = shelleyfee + byronfee
 
       liftIO $ putStrLn $ (show fee :: String) <> " Lovelace"
 
-    UnwitnessedCliFormattedTxBody anyTxBody -> do
-      InAnyShelleyBasedEra sbe txbody <- pure anyTxBody
+    UnwitnessedCliFormattedTxBody (InAnyShelleyBasedEra sbe txbody) -> do
```
Jimbo4350: We  can get rid `UnwitnessedCliFormattedTxBody`

Jimbo4350 (commented):
LGTM. We can now use `calcMinFeeTx` for a more accurate fee calculation.

PR #698, opened 2024-04-05
Fix inversion in --include-stake parser + print "stake" field even when there is no stake (do not hide stake absence)

### LINKED ISSUE #623: cardano-cli `query drep-state` issues with `--include-stake`
```console
$ ./cardano-cli-latest version
cardano-cli 8.20.1.0 - linux-x86_64 - ghc-8.10
git rev 7f49f394d364f8a0b6533051114a63a54cd47a81
```

Doing a query for a specific `--drep-key-hash` does not return the stake even if the flag `--include-stake` is set:

```console
$ ./cardano-cli-latest conway query drep-state --drep-key-hash 7fb3e519ac284cd198e8e403d97b5b3ff58aa798ef85faf39d95971b --include-stake
```
```json
[
    [
        {
            "keyHash": "7fb3e519ac284cd198e8e403d97b5b3ff58aa798ef85faf39d95971b"
        },
        {
            "anchor": null,
            "deposit": 2000000,
            "expiry": 269
        }
    ]
]
```

However, doing a query for `--all-dreps` not only autoincludes(should be off by default) the stake query but when you look at the data for the example drep-key-hash above the output is:
```console
$ ./cardano-cli-latest conway query drep-state --all-dreps
```
```json
...
[
        {
            "keyHash": "7fb3e519ac284cd198e8e403d97b5b3ff58aa798ef85faf39d95971b"
        },
        {
            "anchor": null,
            "deposit": 2000000,
            "expiry": 269,
            "stake": 37992188702
        }
    ],
...
```

@carlhammann @smelc

smelc:
# Changelog

```yaml
- description: |
    Fixed `--include-stake` behavior in `query drep-state`
  type:
```

# Context

Fixes https://github.com/IntersectMBO/cardano-cli/issues/623

# How to trust this PR

I've tested the behavior of this PR on this branch of `cardano-node`: see the HEAD of https://github.com/IntersectMBO/cardano-node/commits/smelc/test-of-cli-pull-698/

# Checklist

- NA New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
@@ -365,13 +365,13 @@ pQueryDRepStateCmd era envCli = do
         <*> pConsensusModeParams
         <*> pNetworkId envCli
         <*> pAllOrOnlyDRepHashSoure
-        <*> Opt.flag WithStake NoStake (mconcat
+        <*> Opt.flag NoStake WithStake (mconcat
               [ Opt.long "include-stake"
               , Opt.help $ mconcat
                  [ "Also return the stake associated with each DRep. "
                  , "The result is the same as with \"drep-stake-distribution\"; "
                  , "this is a convenience option to obtain all information concerning a DRep at once. "
-                 , "This is a potentially expensive query."
+                 , "This is a potentially expensive query, so it's OFF by default."
                  ]
               ]
             )
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
@@ -364,7 +364,7 @@ pQueryDRepStateCmd era envCli = do
         <$> pSocketPath envCli
         <*> pConsensusModeParams
         <*> pNetworkId envCli
-        <*> pAllOrOnlyDRepHashSoure
+        <*> pAllOrOnlyDRepHashSource
         <*> Opt.flag NoStake WithStake (mconcat
               [ Opt.long "include-stake"
               , Opt.help $ mconcat
@@ -394,7 +394,7 @@ pQueryDRepStakeDistributionCmd era envCli = do
       <$> pSocketPath envCli
       <*> pConsensusModeParams
       <*> pNetworkId envCli
-      <*> pAllOrOnlyDRepHashSoure
+      <*> pAllOrOnlyDRepHashSource
       <*> pTarget era
       <*> optional pOutputFile
 
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
@@ -10,7 +10,6 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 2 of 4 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1501,19 +1499,20 @@
...
+      -> (L.Credential L.DRepRole StandardCrypto, L.DRepState StandardCrypto)
+      -> (L.Credential L.DRepRole StandardCrypto, A.Value)
+    drepStateToJson stakeDistr (cred, ds) = (cred, A.object $
+      [ "expiry" .= (ds ^. L.drepExpiryL)
+      , "anchor" .= (ds ^. L.drepAnchorL)
+      , "deposit" .= (ds ^. L.drepDepositL)
+      ] <>
+        (if includeStakeBoolean
+         then [ "stake" .= Map.lookup (L.DRepCredential cred) stakeDistr ]
+         else []))
```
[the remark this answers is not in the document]
smelc: Changed :+1:

cardano-cli/src/Cardano/CLI/EraBased/Options/Query.hs
```diff
@@ -364,14 +364,14 @@ pQueryDRepStateCmd era envCli = do
         <$> pSocketPath envCli
         <*> pConsensusModeParams
         <*> pNetworkId envCli
-        <*> pAllOrOnlyDRepHashSoure
-        <*> Opt.flag WithStake NoStake (mconcat
+        <*> pAllOrOnlyDRepHashSource
+        <*> Opt.flag NoStake WithStake (mconcat
```
smelc, line 368: This is one of the two important bits of this PR: the arguments to [Opt.flag](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative.html#v:flag) are as follows:

1. The default value
2. The modified value

Here the two values were inverted. This change fixes that.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1501,19 +1500,20 @@ runQueryDRepState
...
+      => ToJSON a
+      => Map (L.DRep StandardCrypto) a
+      -> (L.Credential L.DRepRole StandardCrypto, L.DRepState StandardCrypto)
+      -> (L.Credential L.DRepRole StandardCrypto, A.Value)
+    drepStateToJson stakeDistr (cred, ds) = (cred, A.object $
+      [ "expiry" .= (ds ^. L.drepExpiryL)
+      , "anchor" .= (ds ^. L.drepAnchorL)
+      , "deposit" .= (ds ^. L.drepDepositL)
+      ] <>
+        (if includeStakeBoolean
```
smelc: This is the other important bit of this PR: we want to condition the presence of the `stake` field on the value passed to the parser, not on the presence of stake in the data; because if there no stake, the `stake` field will get omitted.

After this PR, if there is no stake delegated to the concerned DRep, the `stake` field will be present and set to `null`; which is a better behavior IMHO.

Jimbo4350: You can `case` on `includeStake` instead of converting to a `Bool`.

Jimbo4350 (approved):
LGTM! One minor comment.

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
smelc: Good point :+1: Done :bow:

PR #723, opened 2024-04-16
Test PR for improving golden CI  (2/2)

smelc:
Test to check that CI will catch golden files that are not being used anymore and should be removed. See e.g.:

image


smelc:
Test finished

PR #730, opened 2024-04-19
CI: add pipeline to create releases and attach binaries to releases

### LINKED ISSUE #725: Create releases on GitHub
To ease usage of `cardano-cli`, we want to create GitHub releases when we tag a commit (see [existing list of tags](https://github.com/IntersectMBO/cardano-cli/tags)). We want to:

* Create a GitHub release, for example [like the node does](https://github.com/IntersectMBO/cardano-node/releases).
* We want the release to contain the binaries for different platforms (like such [a list of assets](https://github.com/jesseduffield/lazygit/releases/tag/v0.40.2))
* We want release building to happen automatically, for example by using a commit's tag as trigger. There should be no manual intervention.

smelc:
# Changelog

```yaml
- description: |
    Add a pipeline named "Release Upload" that attaches binaries of cardano-cli to releases, creating the release if necessary
  type:
```

# Context

* We want to make the life of users easier, so let's give them easier access to binaries :tada: 
* We can generate the release's body automatically, by specifying the changelog; but I plan to do that in a follow-up PR.

Fixes https://github.com/IntersectMBO/cardano-cli/issues/725

# How to trust this PR

The asset attached to [this release](https://github.com/IntersectMBO/cardano-cli/releases/tag/cardano-cli-8.22.0.0) was created by triggering this pipeline manually as follows:

```shell
gh workflow run "Release Upload" -r smelc/automate-shipping-binaries-in-releases -f target_tag=cardano-cli-8.22.0.0
```

Generated asset is this one: 

image

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/.github/workflows/release-upload.yaml
+++ b/.github/workflows/release-upload.yaml
@@ -1,12 +1,37 @@
 name: Release Upload
-# This makes it easy to get download release binaries built using
-# a github action for any tagged commit.
+# Pipeline to download binaries from Hydra builds and create GitHub
+# releases out of them (or attaching the binaries to an already existing release)
 #
-# This workflow builds and uploads the macOS, win64 and linux
-# binary packages as github assets.
+# This pipeline is divided in three jobs:
 #
-# It uses `--builders "" --max-jobs 0` to ensure the assets are
-# from the IOG cache.
+# 1. The "wait_for_hydra" job looks up the status of the Hydra build corresponding to
+#    the tag for which we want to release. If the status is success, the pipeline
+#    continues. If the status cannot be determined or is a failure, the pipeline stops.
+# 2. The "pull" job downloads the binaries from the Hydra build and uploads them as artifacts.
+#    This job uses a matrix to handle all 3 platforms. Uploading as an artifact allows to
+#    do the last job without a matrix.
+#
+#    This job uses `--builders "" --max-jobs 0` to ensure the assets are downloaded
+#    from the IOG cache.
+# 3. The "create_release" job downloads the artifacts and attaches them to the target release.
+#
+# This pipeline is triggered in one of the following 3 ways:
+#
+# 1. When a release is published, this release's commit must have a tag for the pipeline to succeed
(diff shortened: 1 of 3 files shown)


carbolymer:
Do we still need `release` job here https://github.com/IntersectMBO/cardano-api/blob/main/.github/workflows/haskell.yml#L200 after your changes?

smelc:
> Do we still need release job here IntersectMBO/cardano-api@main/.github/workflows/haskell.yml#L200 after your changes?

Indeed no, good catch :+1:

.github/workflows/release-upload.yaml
```diff
@@ -0,0 +1,193 @@
...
+#    do the last job without a matrix.
+#
+#    This job uses `--builders "" --max-jobs 0` to ensure the assets are downloaded
+#    from the IOG cache.
+# 3. The "create_release" job downloads the artifacts and attaches them to the target release.
+#
+# This pipeline is triggered in one of the following 3 ways:
+#
+# 1. When a release is published, this release's commit must have a tag for the pipeline to succeed
+# 2. When a tag is pushed. If this tag has a corresponding release, the release will be augmented.
```
carbolymer, line 21: What happens when you push a tag, then a release gets published and then you remove a tag and push it again (pointing to a different commit)? Will a release get overwritten?

carbolymer (approved):
Nice work!

.github/workflows/release-upload.yaml
smelc: Good question. I don't know and I can't find the answer in the documentation of https://github.com/input-output-hk/action-gh-release. Let me try it out :+1:

smelc: @carbolymer> the release is updated when the tag has changed between two executions of the pipeline :+1: 

See e.g. first release built:

test-tag-1-release-1

and then after the tag has been updated to another commit:

test-tag-1-release-2

carbolymer: Nice, seems fine!

PR #734, opened 2024-04-23
CI: run shellcheck

smelc:
# Changelog

```yaml
- description: |
    CI: add a job to run shellcheck
  type:
```

# Context

While working on https://github.com/IntersectMBO/cardano-cli/pull/733, I witnessed there was no check of the quality of shell scripts in this repo. This PR adds one.

You may think _hey all versioned `*.sh` scripts are being ignored right now_. That is correct: I didn't want to interleave fixing the existing scripts and adding this new check. The point is that this pipeline will automatically start checking new shell scripts, for example the one added by https://github.com/IntersectMBO/cardano-cli/pull/733.

So this pipeline will effectively start checking things when https://github.com/IntersectMBO/cardano-cli/pull/733 is merged. And it will also check future scripts we add (since, by default, we won't think of augmenting `.github/workflows/shellcheck-exceptions.txt` :wink:).

# How to trust this PR

Look at output of recent runs of the new pipeline: https://github.com/IntersectMBO/cardano-cli/actions/workflows/shellcheck.yml

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/.github/workflows/shellcheck-exceptions.txt
+++ b/.github/workflows/shellcheck-exceptions.txt
@@ -1,0 +1,2 @@
+.github/bin/haddocks.sh
+scripts/ci/check-cabal-files.sh
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,39 @@
+name: Shellcheck
+# This pipeline runs shellcheck on all files with extension .sh,
+# except the ones listed in .github/workflows/shellcheck-exceptions.txt.
+#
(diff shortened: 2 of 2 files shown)


carbolymer (approved):
Nice. I think we should add that to cardano-node and cardano-api as well.

smelc:
> Nice. I think we should add that to cardano-node and cardano-api as well.

@carbolymer> sure, will do :+1:

Jimbo4350 (commented):
Surprised this didn't result in some shell script changes!

PR #735, opened 2024-04-23
CI: release Darwin binaries

smelc:
# Changelog

```yaml
- description: |
    CI: release Darwin binaries
  type:
```

# Context

# How to trust this PR

[This pipeline run](https://github.com/IntersectMBO/cardano-cli/actions/runs/8819143561) shows an execution of this new pipeline. It was triggered like this:

```shell
gh workflow run "Release Upload" -r $(git branch --show-current) -f target_tag=1b35eec4cfda3756ba0bca7d6249691bcf7b1829
```

It targets 1b35eec4cfda3756ba0bca7d6249691bcf7b1829, which is one of the only commits that had the Hydra builds running (see https://github.com/IntersectMBO/cardano-cli/pull/686 for more context).

On the considered pipeline run, you can see the new binaries being attached:

image

You can download the binaries and run `file` on them, you'll observe they have the expected architecture:

```shell
> unzip cardano-cli-x86_64-darwin.zip
> file cardano-cli-x86_64-darwin
cardano-cli-x86_64-darwin: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE|HAS_TLV_DESCRIPTORS>
> unzip cardano-cli-aarch64-darwin.zip
> file cardano-cli-aarch64-darwin
cardano-cli-aarch64-darwin: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE|HAS_TLV_DESCRIPTORS>
```

Note that I couldn't test the tail of the pipeline (`create_release` job), because we don't have a tagged commit on which the Hydra Darwin jobs have run (yet). It's fine: we will test it when we tag the next release and I can fix the pipeline at the time, and rerun it manually (possible because it has `workflow_dispatch` and support for targetting a commit different from the one it's being executed on).

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/.github/workflows/release-upload.yaml
+++ b/.github/workflows/release-upload.yaml
@@ -131,9 +131,8 @@ jobs:
     needs: [wait_for_hydra]
     strategy:
       matrix:
-        arch: [linux, x86_64-darwin, aarch64-darwin]
         # TODO generalize
-        # arch: [linux, macos, win64]
+        arch: [x86_64-linux, x86_64-darwin, aarch64-darwin]
     name: "Download Asset"
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-upload.yaml
```diff
@@ -94,29 +96,44 @@ jobs:
...
-              echo "ci/hydra-build:required succeeded"
-              exit 0;;
-            failure)
-              echo "ci/hydra-build:required failed"
-              exit 1;;
-            *)
-              echo "ci/hydra-build:required pending. Waiting 30s..."
-              sleep 30;;
-          esac
+          conclusion=$(gh api "repos/$GITHUB_REPOSITORY/commits/${{ env.TARGET_TAG }}/check-runs" --jq '.check_runs[] | select(.name | test("ci/hydra-build:.*\\.required")) | .conclusion')
```
smelc, line 99: Previously we were matching a single job, whereas now we match 3 of them; so we need a more flexible handling, hence the new lines below.

.github/workflows/release-upload.yaml
```diff
@@ -94,29 +96,44 @@ jobs:
...
+          # could be unknown statuses, which is why we may retry when unsure (see 'sleep') below.
+          echo "ci/hydra-build:.*\\.required returned status: $conclusion"
+          # conclusion is of the form (note the newlines, which matter because we use 'wc -l' below)
+          # success
+          # failure
+          # success
+          # Because we care of the newlines, quoting $conclusion with "" is especially important below!
+          # (see https://stackoverflow.com/questions/22101778/how-to-preserve-line-breaks-when-storing-command-output-to-a-variable)
+
+          # shellcheck disable=SC2126
```
smelc, line 111: shellcheck wants us to use `grep -c` instead of `grep .. | wc -l`. The problem with `grep -c` is that it fails if the number of matches is 0. We don't want that, so we keep the usage of `wc -l`, but need to turn the warning off.

carbolymer (approved):

PR #752, opened 2024-05-07
CI only

smelc:
As I changed the tag of `cardano-cli-8.23.0.0` after the initial PR, and then the tag wasn't the HEAD anymore, hydra didn't run on it. Hence artifacts didn't get created for the tag, and so we can't retrieve them after the fact (to recreate the release, which I deleted manually when testing the PATCH endpoint - this is another story).

So this PR retriggers hydra on the cardano-cli-8.23.0.0 tag.


smelc:
Closing as I could have https://github.com/IntersectMBO/cardano-cli/releases/tag/cardano-cli-8.23.0.0 created, thanks to the runs of this PR.

PR #778, opened 2024-05-27
Add --current-treasury-value and --treasury-donation to transaction build and friends

smelc:
# Changelog

```yaml
- description: |
    Add --current-treasury-value and --treasury-donation to transaction build and friends
# uncomment types applicable to the change:
  type:
  - feature        # introduces a new feature
  - breaking       # the API has changed in a breaking way
  # - compatible     # the API has changed but is non-breaking
  # - optimisation   # measurable performance improvements
  # - refactoring    # QoL changes
  # - bugfix         # fixes a defect
  # - test           # fixes/modifies tests
  # - maintenance    # not directly related to the code
  # - release        # related to a new release preparation
  # - documentation  # change in code docs, haddocks...
```

# Context

Fixes https://github.com/IntersectMBO/cardano-cli/issues/590

# How to trust this PR

We tested this new feature in `cardano-testnet` in this `cardano-node` PR: https://github.com/IntersectMBO/cardano-node/pull/5866

# Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages
- [x] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [x] Self-reviewed the diff
- [x] Consider https://github.com/IntersectMBO/cardano-cli/pull/783#discussion_r1638321013
- [x] Consider https://github.com/IntersectMBO/cardano-cli/pull/783#discussion_r1638321175
- [x] Use `CurrentTreasuryValue` type from API instead of introducing a new one (see how API PR is received)

PR #783, opened 2024-06-07
Transaction.hs: reduce boilerplate and move towards ShelleyBasedEra

smelc:
# Changelog

```yaml
- description: |
    Transaction.hs: reduce boilerplate and move towards ShelleyBasedEra
  type:
```

# How to trust this PR

* It doesn't actually change the runtime behavior
* It's more deletion that additions :1st_place_medal: 

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
@@ -68,7 +68,7 @@ import           Data.Function ((&))
 import qualified Data.List as List
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (catMaybes, fromMaybe)
+import           Data.Maybe (catMaybes, fromMaybe, mapMaybe)
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import qualified Data.Text as Text
@@ -644,8 +644,7 @@ constructTxBodyContent sbe mScriptValidity mPparams inputsAndMaybeScriptWits rea
                        reqSigners fee txAuxScripts txMetadata txUpdateProposal
                        votingProcedures proposals
                        = do
-  let era = toCardanoEra sbe -- TODO: Propagate SBE
-      allReferenceInputs = getAllReferenceInputs
+  let allReferenceInputs = getAllReferenceInputs
                              inputsAndMaybeScriptWits
                              (snd valuesWithScriptWits)
                              certsAndMaybeScriptWits
@@ -656,20 +655,14 @@ constructTxBodyContent sbe mScriptValidity mPparams inputsAndMaybeScriptWits rea
 
   validatedCollateralTxIns <- validateTxInsCollateral sbe txinsc
   validatedRefInputs <- validateTxInsReference sbe allReferenceInputs
-  validatedTotCollateral
-    <- first TxCmdTotalCollateralValidationError $ validateTxTotalCollateral era mTotCollateral
-  validatedRetCol
-    <- first TxCmdReturnCollateralValidationError $ validateTxReturnCollateral era mReturnCollateral
+  validatedTotCollateral <- first TxCmdNotSupportedInAnyCardanoEraValidationError $ validateTxTotalCollateral sbe mTotCollateral
+  validatedRetCol        <- first TxCmdNotSupportedInAnyCardanoEraValidationError $ validateTxReturnCollateral sbe mReturnCollateral
   let txFee = TxFeeExplicit sbe fee
-  validatedLowerBound <- first TxCmdTxValidityLowerBoundValidationError (validateTxValidityLowerBound era mLowerBound)
-  validatedReqSigners <- first TxCmdRequiredSignersValidationError $ validateRequiredSigners era reqSigners
-  validatedTxWtdrwls <- first TxCmdTxWithdrawalsValidationError $ validateTxWithdrawals era withdrawals
-  validatedTxCerts <- first TxCmdTxCertificatesValidationError $ validateTxCertificates era certsAndMaybeScriptWits
+  validatedLowerBound <- first TxCmdNotSupportedInAnyCardanoEraValidationError $ validateTxValidityLowerBound sbe mLowerBound
+  validatedReqSigners <- first TxCmdNotSupportedInAnyCardanoEraValidationError $ validateRequiredSigners sbe reqSigners
   validatedMintValue <- createTxMintValue sbe valuesWithScriptWits
-  validatedTxScriptValidity <- first TxCmdScriptValidityValidationError $ validateTxScriptValidity era mScriptValidity
+  validatedTxScriptValidity <- first TxCmdNotSupportedInAnyCardanoEraValidationError $ validateTxScriptValidity sbe mScriptValidity
   validatedVotingProcedures <- first TxCmdTxGovDuplicateVotes $ convertToTxVotingProcedures votingProcedures
-  validatedPParams <- first TxCmdProtocolParametersValidationError
-                                  $ validateProtocolParameters era (LedgerProtocolParameters <$> mPparams)
   return $ shelleyBasedEraConstraints sbe $ (defaultTxBodyContent sbe
              & setTxIns (validateTxIns inputsAndMaybeScriptWits)
              & setTxInsCollateral validatedCollateralTxIns
@@ -683,20 +676,24 @@ constructTxBodyContent sbe mScriptValidity mPparams inputsAndMaybeScriptWits rea
              & setTxMetadata txMetadata
              & setTxAuxScripts txAuxScripts
              & setTxExtraKeyWits validatedReqSigners
-             & setTxProtocolParams validatedPParams
-             & setTxWithdrawals validatedTxWtdrwls
-             & setTxCertificates validatedTxCerts
+             & setTxProtocolParams (BuildTxWith $ LedgerProtocolParameters <$> mPparams)
+             & setTxWithdrawals (TxWithdrawals sbe $ map convertWithdrawals withdrawals)
+             & setTxCertificates (convertCertificates sbe certsAndMaybeScriptWits)
              & setTxUpdateProposal txUpdateProposal
              & setTxMintValue validatedMintValue
              & setTxScriptValidity validatedTxScriptValidity)
              -- TODO: Create set* function for proposal procedures and voting procedures
-             { txProposalProcedures = forEraInEonMaybe era (`Featured` convToTxProposalProcedures proposals)
-             , txVotingProcedures = forEraInEonMaybe era (`Featured` validatedVotingProcedures)
+             { txProposalProcedures = forShelleyBasedEraInEonMaybe sbe (`Featured` convToTxProposalProcedures proposals)
+             , txVotingProcedures = forShelleyBasedEraInEonMaybe sbe (`Featured` validatedVotingProcedures)
              }
-
-
-
-
+ where
+  convertWithdrawals
+    :: (StakeAddress, L.Coin, Maybe (ScriptWitness WitCtxStake era))
+    -> (StakeAddress, L.Coin, BuildTxWith BuildTx (Witness WitCtxStake era))
+  convertWithdrawals (sAddr, ll, mScriptWitnessFiles) =
+    case mScriptWitnessFiles of
+      Just sWit -> (sAddr, ll, BuildTxWith $ ScriptWitness ScriptWitnessForStakeAddr sWit)
+      Nothing   -> (sAddr, ll, BuildTxWith $ KeyWitness KeyWitnessForStakeAddr)
 
 runTxBuild :: ()
   => ShelleyBasedEra era
@@ -773,13 +770,8 @@ runTxBuild
   Refl <- testEquality era nodeEra
     & hoistMaybe (TxCmdTxNodeEraMismatchError $ NodeEraMismatchError era nodeEra)
 
-  validatedTxCerts
-     <- hoistEither
-          . first TxCmdTxCertificatesValidationError
-          $ validateTxCertificates era certsAndMaybeScriptWits
-
   let certs =
-        case validatedTxCerts of
+        case convertCertificates sbe certsAndMaybeScriptWits of
           TxCertificates _ cs _ -> cs
           _ -> []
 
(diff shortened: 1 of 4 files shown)


cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -240,25 +187,16 @@ validateTxWithdrawals era withdrawals = do
...
-
-instance Error TxCertificatesValidationError where
-  prettyError (TxCertificatesValidationNotSupported e) =
-    "Transaction certificates are not supported in " <> pretty e
-
 -- TODO: Because we have separated Byron related transaction
--- construction into separate commands, we can parameterize this
--- on ShelleyBasedEra era and remove Either TxCertificatesValidationError
+-- construction into separate commands, we can remove
+-- Either TxNotSupportedInEraValidationError
```
smelc: Does this comment mean I should just remove the call to `conjureWitness`?

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -70,64 +65,80 @@ validateScriptSupportedInEra era script@(ScriptInAnyLang lang _) =
...
-instance Error TxTotalCollateralValidationError where
-  prettyError (TxTotalCollateralNotSupported era) =
-    "Transaction collateral not supported in " <> pretty era
-
-validateTxTotalCollateral :: CardanoEra era
+data TxNotSupportedInEraValidationError era =
+    -- | First argument is the kind of data that is not supported.
+    -- Second argument in the era that doesn't support the data.
+    TxNotSupportedInAnyCardanoEraValidationError T.Text AnyCardanoEra
+    -- | First argument is the kind of data that is not supported.
```
carbolymer: I'm not sure if this haddock is 100% correct - I think it should be after `|`, as it references the second constructor.

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
carbolymer: Well, `sbe == supported` in the function body, so this `conjureWitness` will never fail, so i'd say yes. The same for other places in this PR, where we changed `CardanoEra` to `ShelleyBasedEra`.

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -70,64 +65,80 @@ validateScriptSupportedInEra era script@(ScriptInAnyLang lang _) =
...
-
-validateTxReturnCollateral :: CardanoEra era
+validateTxCurrentTreasuryValue :: ()
+  => ShelleyBasedEra era
+  -> Maybe TxCurrentTreasuryValue
+  -> Either (TxNotSupportedInEraValidationError era) (Maybe (Featured ConwayEraOnwards era L.Coin))
+validateTxCurrentTreasuryValue sbe mCurrentTreasuryValue =
+  case mCurrentTreasuryValue of
+    Nothing -> Right Nothing
+    Just (TxCurrentTreasuryValue { unTxCurrentTreasuryValue }) ->
```
carbolymer: suggestion:
```suggestion
validateTxCurrentTreasuryValue sbe =
  mapM $ \(TxCurrentTreasuryValue { unTxCurrentTreasuryValue }) ->
```
or you can use `forM`, ...or neither 😄

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -70,64 +65,80 @@ validateScriptSupportedInEra era script@(ScriptInAnyLang lang _) =
...
+      caseShelleyToBabbageOrConwayEraOnwards
+        (const $ Left $ TxNotSupportedInShelleyBasedEraValidationError "Current treasury value" sbe)
+        (\cOnwards -> Right $ Just $ Featured cOnwards unTxCurrentTreasuryValue)
+        sbe
+
+validateTxTreasuryDonation :: ()
+  => ShelleyBasedEra era
+  -> Maybe TxTreasuryDonation
+  -> Either (TxNotSupportedInEraValidationError era) (Maybe (Featured ConwayEraOnwards era L.Coin))
+validateTxTreasuryDonation sbe mTreasuryDonation =
```
carbolymer: same suggestion as above

carbolymer (approved):
Good job!

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -272,57 +212,33 @@ validateTxCertificates era certsAndScriptWitnesses = cardanoEraConstraints era $
...
-newtype TxProtocolParametersValidationError
-  = ProtocolParametersNotSupported AnyCardanoEra
-  deriving Show
-
-instance Error TxProtocolParametersValidationError where
-  prettyError (ProtocolParametersNotSupported e) =
-    "Transaction protocol parameters are not supported in " <> pretty e
-
 validateProtocolParameters
   :: CardanoEra era
```
Jimbo4350: This will only fail if we call `validateProtocolParameters` with `ByronEra`. However the code won't compile if we call this with `ByronEra`:

```
ghci> validateProtocolParameters ByronEra Nothing

<interactive>:22:1: error: [GHC-39999]
    • No instance for ‘IsShelleyBasedEra ByronEra’
        arising from a use of ‘print’
    • In a stmt of an interactive GHCi command: print it
```

What we should do is change the parameter from `CardanoEra era` to `ShelleyBasedEra era` but actually in this case no validation is done! 

So we can remove this function and directly use `setTxProtocolParams`.

Have a look at the other functions that demand `CardanoEra era` but should be parameterized on `ShelleyBasedEra era`. If no validation is done, we can just drop them. I'd do that in a separate PR.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
smelc: Since I rebased this PR on `main` (instead of https://github.com/IntersectMBO/cardano-cli/pull/778) this is irrelevant now, but I'll keep it mind when revisiting https://github.com/IntersectMBO/cardano-cli/pull/778 :+1:

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
smelc: Since I rebased this PR on `main` (instead of https://github.com/IntersectMBO/cardano-cli/pull/778) this is irrelevant now, but I'll keep it mind when revisiting https://github.com/IntersectMBO/cardano-cli/pull/778 :+1:

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
smelc: OK will do in a follow-up PR right away :+1: 

@Jimbo4350> so can you approve in the meantime?

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
smelc: @carbolymer> nah it's all good: 

image

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
Jimbo4350: I can't because it would be adding churn. If you do what I explained above it should also result in the removal of the unnecessary error constructors.

smelc: @Jimbo4350> done :+1: Please rereview.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -656,20 +655,15 @@ constructTxBodyContent sbe mScriptValidity mPparams inputsAndMaybeScriptWits rea
...
+  validatedTotCollateral <- first TxCmdNotSupportedInEraValidationError $ validateTxTotalCollateral sbe mTotCollateral
+  validatedRetCol        <- first TxCmdNotSupportedInEraValidationError $ validateTxReturnCollateral sbe mReturnCollateral
   let txFee = TxFeeExplicit sbe fee
-  validatedLowerBound <- first TxCmdTxValidityLowerBoundValidationError (validateTxValidityLowerBound era mLowerBound)
-  validatedReqSigners <- first TxCmdRequiredSignersValidationError $ validateRequiredSigners era reqSigners
-  validatedTxWtdrwls <- first TxCmdTxWithdrawalsValidationError $ validateTxWithdrawals era withdrawals
-  validatedTxCerts <- first TxCmdTxCertificatesValidationError $ validateTxCertificates era certsAndMaybeScriptWits
+  validatedLowerBound <- first TxCmdNotSupportedInEraValidationError $ validateTxValidityLowerBound sbe mLowerBound
+  validatedReqSigners <- first TxCmdNotSupportedInEraValidationError $ validateRequiredSigners sbe reqSigners
+  -- validatedTxCerts    <- first TxCmdNotSupportedInEraValidationError $ validateTxCertificates sbe certsAndMaybeScriptWits
```
Jimbo4350: Left over comment?

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -656,20 +655,15 @@ constructTxBodyContent sbe mScriptValidity mPparams inputsAndMaybeScriptWits rea
...
-  validatedRetCol
-    <- first TxCmdReturnCollateralValidationError $ validateTxReturnCollateral era mReturnCollateral
+  validatedTotCollateral <- first TxCmdNotSupportedInEraValidationError $ validateTxTotalCollateral sbe mTotCollateral
+  validatedRetCol        <- first TxCmdNotSupportedInEraValidationError $ validateTxReturnCollateral sbe mReturnCollateral
   let txFee = TxFeeExplicit sbe fee
-  validatedLowerBound <- first TxCmdTxValidityLowerBoundValidationError (validateTxValidityLowerBound era mLowerBound)
-  validatedReqSigners <- first TxCmdRequiredSignersValidationError $ validateRequiredSigners era reqSigners
-  validatedTxWtdrwls <- first TxCmdTxWithdrawalsValidationError $ validateTxWithdrawals era withdrawals
-  validatedTxCerts <- first TxCmdTxCertificatesValidationError $ validateTxCertificates era certsAndMaybeScriptWits
+  validatedLowerBound <- first TxCmdNotSupportedInEraValidationError $ validateTxValidityLowerBound sbe mLowerBound
```
Jimbo4350: Have a look at `validateTxValidityLowerBound` and the actual data type in `cardano-api`:
```
data TxValidityLowerBound era where

  TxValidityNoLowerBound
    :: TxValidityLowerBound era

  TxValidityLowerBound
    :: AllegraEraOnwards era
    -> SlotNo
    -> TxValidityLowerBound era
```
`validateTxValidityLowerBound` should be updated to reflect the type. Using `conjureWitness` is misleading because the type signature of `validateTxValidityLowerBound` gives the impression that any `CardanoEra era` is supported but this is not true.

Have a look at all of the validation functions with this in mind and update the type signature accordingly.

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
smelc: Good catch, removed :+1:

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
smelc: @Jimbo4350> not sure what you want here. In the current state of this PR, `validateTxValidityLowerBound` takes a `ShelleyBasedEra era` witness (this is an improvement, it was taking `CardanoEra era` before this PR).

If I tighten its type to mimick the API one, for example requiring `AllegraEraOnwards era`, we are going to push the era's verification to the caller (i.e. in `Transaction.hs`) and I think the point of the current design was to have verifications done in `TxValidationErrors.hs`.

What I could do is to remove the call to `conjureWitness` and use [caseShelleyEraOnlyOrAllegraEraOnwards](https://github.com/IntersectMBO/cardano-api/blob/smelc/more-logging-in-evaluateTransactionExecutionUnit/cardano-api/internal/Cardano/Api/Eras/Case.hs#L80). This would make the code more idiomatic, since that's what we do in more modern parts of the codebase. Is that the change you want?

Jimbo4350: I was on the wrong branch :man_facepalming:. This is fine as is.

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
```diff
@@ -70,64 +59,54 @@ validateScriptSupportedInEra era script@(ScriptInAnyLang lang _) =
...
-
-instance Error TxTotalCollateralValidationError where
-  prettyError (TxTotalCollateralNotSupported era) =
-    "Transaction collateral not supported in " <> pretty era
-
-validateTxTotalCollateral :: CardanoEra era
+data TxNotSupportedInEraValidationError era =
+    -- | First argument is the kind of data that is not supported.
+    -- Second argument is the era that doesn't support the data.
+    TxNotSupportedInAnyCardanoEraValidationError T.Text AnyCardanoEra
```
Jimbo4350, line 65: This constructor can actually be removed.

Jimbo4350 (approved):
LGTM! However you can remove the `TxNotSupportedInAnyCardanoEraValidationError` constructor.

cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs
smelc: Good catch, done :+1:

Jimbo4350: It should return `AnyShelleyBasedEra` because: https://github.com/IntersectMBO/cardano-cli/pull/783/files#diff-a38343c1825e6e334224d0a8a1ba78a458652970f00c54f5fb0d040be28f822eR606

smelc: @Jimbo4350> not really, because this constructor is passed as a function to [conjureWitness](https://github.com/IntersectMBO/cardano-cli/blob/smelc/reduce-boilerplate-modernize-TxValidationErrors/cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs#L151) and there remains [one single caller](https://github.com/IntersectMBO/cardano-cli/blob/smelc/reduce-boilerplate-modernize-TxValidationErrors/cardano-cli/src/Cardano/CLI/Types/Errors/TxValidationError.hs#L146) to `conjureWitness` that uses a `CardanoEra era` value.

Once this last caller is eliminated, we will be able to tighten `TxNotSupportedInAnyCardanoEraValidationError`'s type.

Jimbo4350: We could use `shelleyBasedToCardanoEra`.

smelc: @Jimbo4350> but this is awkward because `conjureWitness` expects a function to build the error, to which it applies the `CardanoEra`. If I want to keep `conjureWitness` existing type (which I think is nice), I need to do something like this:

image

which is weird, because the value passed by `conjureWitness` gets ignored when building the error.

PR #798, opened 2024-06-21
Add Haskell Language Server check to CI

palas:
This PR adds a check to the CI that ensures that it works with the developer shell as it is.

# Changelog

```yaml
- description: |
    Add CI check that ensures HLS works with developer shell
  type:
  - test
```

# Context

Recently we had an issue with HLS not working because of a bug. We could have catched that if we had a test in the CI that ensured HLS worked. This PR aims to address this.

# How to trust this PR

Check the workflow makes sense. But basically, I would just look at [the CI's output](https://github.com/IntersectMBO/cardano-cli/actions/runs/9618399739/job/26532205253?pr=798).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -10,20 +10,42 @@ permissions:
   contents: read
   
 jobs:
+
   test-hls-works:
+    env:
+      # Modify this value to "invalidate" the cache.
+      HLS_CACHE_VERSION: "2024-06-25"
+  
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - name: Check out repo
+        uses: actions/checkout@v4
       - name: Install Nix with good defaults
         uses: input-output-hk/install-nix-action@v20
         with:
           extra_nix_config: |
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
             substituters = https://cache.iog.io/ https://cache.nixos.org/
-          nix_path: nixpkgs=channel:nixos-unstable
-      - uses: cachix/install-nix-action@v18
+          nix_path: nixpkgs=channel:nixos-unstable 
+      - name: Open nix environment
+        uses: rrbutani/use-nix-shell-action@v1
+      - name: Update dependencies
+        run: cabal update; cabal freeze
+      - name: Obtain GHC version
+        run: |
+          echo "VERSION=$(ghc --numeric-version)" >> "$GITHUB_OUTPUT"
+        id: ghc
+      - name: HLS caching
+        uses: actions/cache@v4
         with:
-          nix_path: nixpkgs=channel:nixos-unstable
-      # Make the Nix environment available to next steps
-      - uses: rrbutani/use-nix-shell-action@v1
-      - run: cabal update; haskell-language-server
+          path: |
+            /home/runner/.cache/hie-bios
+            /home/runner/.cache/ghcide
+            /home/runner/.local/state/cabal
+            .ghc.environment.x86_64-linux-9.8.2
+            dist-newstyle
+          key: hls-cache-${{ env.HLS_CACHE_VERSION }}-${{ runner.os }}-${{ steps.ghc.outputs.VERSION }}-${{ hashFiles('**/cabal.project.freeze') }}-${{ hashFiles('**/*.cabal', '**/cabal.project') }}
+          restore-keys: |
+            hls-cache-${{ env.HLS_CACHE_VERSION }}-${{ runner.os }}-${{ steps.ghc.outputs.VERSION }}-${{ hashFiles('**/cabal.project.freeze') }}-
+      - name: Test HLS works
+        run: haskell-language-server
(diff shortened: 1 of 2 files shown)


.github/workflows/hls.yml
```diff
@@ -0,0 +1,19 @@
...
+  
+jobs:
+  test-hls-works:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+    - uses: cachix/install-nix-action@daddc62a2e67d1decb56e028c9fa68344b9b7c2a
+      with:
+        nix_path: nixpkgs=channel:nixos-23.11
+    - run: nix develop --accept-flake-config --command bash -c "cabal update; haskell-language-server"
```
smelc:
```suggestion
      - uses: actions/checkout@v4
      - name: Install Nix with good defaults
        uses: input-output-hk/install-nix-action@v20
        with:
          extra_nix_config: |
            trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
            substituters = https://cache.iog.io/ https://cache.nixos.org/
          nix_path: nixpkgs=channel:nixos-unstable
      - uses: cachix/install-nix-action@v18
        with:
          nix_path: nixpkgs=channel:nixos-unstable
      # Make the Nix environment available to next steps
      - uses: rrbutani/use-nix-shell-action@v1
      - run: haskell-language-server
```

Rationale: use same setup as other workflows

palas: Solved in: https://github.com/IntersectMBO/cardano-cli/pull/798/commits/05d4c08d687cb54423fec424a66f0db3312501d7
I had to add `cabal update` to the `run` bit too.

smelc (approved):

carbolymer (changes requested):
I'm fine with the version bump. This GHA however takes 20 minutes, which is too long of a wait for each PR to get merged.

I don't remember, can you confirm that each PR will wait for this new check to complete before merging?

smelc:
@carbolymer> right now the new check is not _required_ so we can merge manually before this new check is finished.

Regarding the merge queue's behavior, [the documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) says it waits for required checks to be satisfied. So not sure if it merges as soon as _required_ checks are green, ignoring the others. I'd be surprised if it was the case.

.github/workflows/hls.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: cachix/install-nix-action@v18
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - name: Open nix environment
+        uses: rrbutani/use-nix-shell-action@v1
+      - name: Update dependencies
+        run: cabal update; cabal freeze
+      - name: Obtain GHC version
+        run: |
+          echo "VERSION=$(ghc --version | grep -oE '[^ ]+$')" >> $GITHUB_OUTPUT
```
smelc:
```suggestion
          echo "VERSION=$(ghc --numeric-version)" >> $GITHUB_OUTPUT
```

.github/workflows/hls.yml
```diff
@@ -0,0 +1,47 @@
...
+        id: ghc
+      - name: Cache hie-bios and ghcide 
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/hie-bios
+            ~/.cache/ghcide
+          key: hls-cache-${{ runner.os }}-${{ steps.ghc.outputs.VERSION }}-${{ hashFiles('**/cabal.project.freeze') }}-${{ hashFiles('**/*.cabal', '**/cabal.project') }}
```
smelc: Also maybe include a mechanism similar to `CABAL_CACHE_VERSION` from the CLI? I'm pretty confident at some point the cache will get wrong/corrupted :upside_down_face:

.github/workflows/hls.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Open nix environment
+        uses: rrbutani/use-nix-shell-action@v1
+      - name: Update dependencies
+        run: cabal update; cabal freeze
+      - name: Obtain GHC version
+        run: |
+          echo "VERSION=$(ghc --version | grep -oE '[^ ]+$')" >> $GITHUB_OUTPUT
+        id: ghc
+      - name: Cache hie-bios and ghcide 
+        uses: actions/cache@v3
```
smelc:
```suggestion
        uses: actions/cache@v4
```

Why not use the latest version available?

.github/workflows/hls.yml
palas: Oh, that is convenient

.github/workflows/hls.yml
palas: I was worrying about the same, I will ~include the same variable here~ do the same thing here

palas:
> I'm fine with the version bump. This GHA however takes 20 minutes, which is too long of a wait for each PR to get merged.
> 
> I don't remember, can you confirm that each PR will wait for this new check to complete before merging?

It takes 4 min now on cache hit. And, it is my understanding, that it should hit cache as long as dependencies for the project don't change

hls-cached

carbolymer (approved):
If it takes 4 mins, I'm fine with it! 🚀

PR #826, opened 2024-07-05
transaction-build and build-estimate: include current treasury value only if a donation is being done

### LINKED ISSUE #825: Build cmd is populating currentTreasuryValue on every TX. It should not.
## Description

 In Conway era, CLI build is populating the field `current treasury value` on every transaction body. This is problematic. Its causes Plutus V1/V2  transactions to fail with `CurrentTreasuryFieldNotSupported` . Ledger expects that field to be empty on such transactions. it can only be not null on PlutusV3. 

It can also be problematic at epoch boundaries for normal transactions. i.e. when the transaction is generated on one epoch and submitted on the next.

See https://github.com/IntersectMBO/cardano-ledger/issues/4470
and https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1720120358499899

...

## Possible Solution

We should only populate this field when it is strictly needed. i.e. only on treasury donations (I think that is the only time the field is needed).

Have users do a two step process for this query treasury -> build so that the field is only used when strictly necesary and left empty most of the time.

smelc:
# Changelog

```yaml
- description: |
    transaction-build and build-estimate: include current treasury value only if a donation is being done
  type:
```

# Context

Fixes https://github.com/IntersectMBO/cardano-cli/issues/825

# How to trust this PR

* A new test is added
* Maybe add a negative test?

# Checklist

- [ ] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -206,6 +206,12 @@ runTransactionBuildCmd
       & onLeft (left . TxCmdQueryConvenienceError . AcqFailure)
       & onLeft (left . TxCmdQueryConvenienceError)
 
+  let currentTreasuryValueAndDonation =
+        case (treasuryDonation, unFeatured <$> featuredCurrentTreasuryValueM) of
+          (Nothing, _) -> Nothing -- We shouldn't specify the treasury value when no donation is being done
+          (Just _td, Nothing) -> undefined -- TODO: Current treasury value couldn't be obtained but is required: we should fail suggesting that the node's version is too old
```
Jimbo4350: Left over `undefined`

Jimbo4350 (commented):
There is a left over `undefined`.

Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -255,7 +261,7 @@ runTransactionBuildEstimateCmd
   :: ()
   => Cmd.TransactionBuildEstimateCmdArgs era
   -> ExceptT TxCmdError IO ()
-runTransactionBuildEstimateCmd
+runTransactionBuildEstimateCmd -- TODO change type
```
smelc, line 264: This is a leftover TODO. It can be removed. cc @CarlosLopezDeLara

PR #874, opened 2024-08-23
create-testnet-data: better behavior for create-testnet-data --total-supply and --delegated-supply

smelc:
# Changelog

```yaml
- description: |
    create-testnet-data's --total-supply option doesn't have a default anymore. The default value is to take the value from the shelley genesis file (if provided, otherwise this file is defaulted, so total supply comes from the default shelley genesis). create-testnet-data's --delegated-supply option doesn't have a default anymore. The default is to use half of the total supply.
  type:
```

# Context

Contributes to fixing https://github.com/IntersectMBO/cardano-node/issues/5953

The API for `create-testnet-data`'s total supply and delegated supply was bad:

1. `--total-supply` had a default value. So we would never pick up the value coming from the shelley genesis file. Now the behavior is that, if `--total-supply` is specified, it overrides the value from the shelley genesis file. Because the shelley genesis file is defaulted, there is always a value available.
2. `--delegated-supply` had a default value. This does not play well with the value of total supply coming from either the genesis file or `--total-supply`. Now the default is to take half of the total supply (no matter from where it comes from).

# How to trust this PR

* Tested with `cardano-testnet` at fecf2bf0c728060c0ff408d2a7fecc152c25dd4f

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
+++ b/cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
@@ -1,15 +1,16 @@
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 
 module Test.Golden.CreateTestnetData where
 
+import           Cardano.Api
 import           Cardano.Api.Ledger (ConwayGenesis (..), StandardCrypto)
 import           Cardano.Api.Shelley (ShelleyGenesis (..))
 
 import qualified Cardano.Ledger.Shelley.API as L
 
 import           Control.Monad
-import           Control.Monad.IO.Class
 import           Data.List (intercalate, sort)
 import qualified Data.Sequence.Strict as Seq
 import           Data.Word (Word32)
@@ -18,6 +19,7 @@ import           System.Directory
 import           System.Directory.Extra (listDirectories)
 import           System.FilePath
 
+import           Test.Cardano.CLI.Aeson
 import           Test.Cardano.CLI.Util (FileSem, bracketSem, execCardanoCLI, newFileSem)
 
 import           Hedgehog (Property)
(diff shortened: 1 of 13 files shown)


cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
```diff
@@ -191,3 +191,37 @@ hprop_golden_create_testnet_data_deleg_non_deleg =
...
+        , tweakedShelleyGenesisFp
+        , "--out-dir"
+        , outputDir
+        ]
+
+    outputGenesis :: ShelleyGenesis StandardCrypto <-
+      H.readJsonFileOk $ outputDir </> "shelley-genesis.json"
+
+    -- Because we don't test this elsewhere in this file:
+    (L.sgMaxLovelaceSupply outputGenesis) H.=== tweakedValue
```
Jimbo4350: Rather than checking a single value, why not do a golden test on the entire genesis value?

smelc: @Jimbo4350> I hadn't done it initially, because the generated genesis file is not stable.

However, I brought [cardano-node's redactJsonFieldsInFile](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Query.hs#L428) function to redact the unstable fields, and I now made this a golden test.

smelc: @Jimbo4350> I did that in an additional commit (that I will fixup later on), to make your review easier.

cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
```diff
@@ -191,3 +193,43 @@ hprop_golden_create_testnet_data_deleg_non_deleg =
...
+    let tweakedValue = 3_123_456_000_000
+        tweakedShelleyGenesis = vanillaShelleyGenesis{L.sgMaxLovelaceSupply = tweakedValue}
+        tweakedShelleyGenesisFp = tempDir </> "tweaked-shelley-genesis.json"
+
+    void $ liftIO $ writeFileJSON tweakedShelleyGenesisFp tweakedShelleyGenesis
+
+    let outputDir = tempDir </> "out"
+    void $
+      execCardanoCLI
+        [ "conway"
```
Jimbo4350, line 216: We shouldn't hardcore the era here. Use `eraToString`.

cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
```diff
@@ -191,3 +193,43 @@ hprop_golden_create_testnet_data_deleg_non_deleg =
     -- https://github.com/IntersectMBO/cardano-cli/issues/631 witnesses.
 
     H.assertWith (fromIntegral onlyHolderCoin) (\ohc -> ohc > totalSupply `div` 2)
+
+-- Execute this test with:
+-- @cabal test cardano-cli-golden --test-options '-p "/golden create testnet data total supply template/"'@
+--
+-- This test checks that we use the total supply value from the shelley template when --total-supply
+-- is not specified. It was broken in the past (see https://github.com/IntersectMBO/cardano-node/issues/5953).
+hprop_golden_create_testnet_data_total_supply_template :: Property
```
Jimbo4350: The property name should be updated. It's a golden test of the output of `create-testnet-data`.

Jimbo4350 (approved):
LGTM!

cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
smelc: Done :+1:

cardano-cli/test/cardano-cli-golden/Test/Golden/CreateTestnetData.hs
smelc: Actually there is no era witness floating around here. This is the one place where we define the era. (and by the way, `eraToString` is not a thing in CLI nor API, it's a `cardano-testnet` thing)(which maybe we should move to API?).

So I'm merging like this.

PR #918, opened 2024-09-30
Updates cardano-cli to not use PReferenceScript/SReferenceScript ScriptHash field

Swordlash:
# Changelog

```yaml
- description: Updates cardano-cli to not use PReferenceScript/SReferenceScript ScriptHash field
  type:
  - compatible
  - refactoring
```

# Context

See the [upstream PR](https://github.com/IntersectMBO/cardano-api/pull/642).
The `Maybe ScriptHash` field in `P/SReferenceScript` is redundant in `cardano-api`, causing confusion. Nonetheless it's used in `cardano-cli` as a temporary container for user-provided input. This PR refactors it out so the field can be safely removed from upstream.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
@@ -204,7 +204,7 @@ runTransactionBuildCmd
     let allReferenceInputs =
           getAllReferenceInputs
             inputsAndMaybeScriptWits
-            (snd valuesWithScriptWits)
+            (snd <$> snd valuesWithScriptWits)
             certsAndMaybeScriptWits
             withdrawalsAndMaybeScriptWits
             votingProceduresAndMaybeScriptWits
@@ -697,7 +697,7 @@ runTxBuildRaw
   -- ^ Tx upper bound
   -> Lovelace
   -- ^ Tx fee
-  -> (Value, [ScriptWitness WitCtxMint era])
+  -> (Value, [(Maybe PolicyId, ScriptWitness WitCtxMint era)])
   -- ^ Multi-Asset value(s)
   -> [(Certificate era, Maybe (ScriptWitness WitCtxStake era))]
   -- ^ Certificate with potential script witness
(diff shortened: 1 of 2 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
```diff
@@ -1354,11 +1354,11 @@ createTxMintValue era (val, scriptWitnesses) =
...
  where
   gatherMintingWitnesses
-    :: [ScriptWitness WitCtxMint era]
+    :: [(Maybe PolicyId, ScriptWitness WitCtxMint era)]
     -> [(PolicyId, ScriptWitness WitCtxMint era)]
   gatherMintingWitnesses [] = []
-  gatherMintingWitnesses (sWit : rest) =
-    case scriptWitnessPolicyId sWit of
+  gatherMintingWitnesses ((pid'm, sWit) : rest) =
+    case scriptWitnessPolicyId pid'm sWit of
```
smelc:
```suggestion
  gatherMintingWitnesses ((mPid, sWit) : rest) =
    case scriptWitnessPolicyId mPid sWit of
```

Rationale: having a quote in the middle of a variable name is surprising

smelc:
@Swordlash> please format your changes with:

```
fourmolu -i ...modified hs files...
stylish-haskell -i ...modified hs files...
```

(in this order: `fourmolu` first, then `stylish-haskell`)

cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs
Swordlash: I picked up this style in Standard Chartered (`'m` for `Maybe`, `'e` for `Either`). No issue, I will adhere to suggested style.

Swordlash: All updated, squashed and formatted.

Jimbo4350:
@Swordlash thank you for bringing this to our attention. I have been aware of this for some time but haven't been able to dedicate the time to a solution. We're discussing a potential solution that involves modifying the `cardano-api` and avoids introducing `Maybe PolicyId` as it's done here. I've  asked @carbolymer to work on it.

Jimbo4350 (changes requested):
We are discussing another way to modify `cardano-api` to avoid having to specify `Maybe PolicyId`.

Swordlash:
Thanks @Jimbo4350, lmk if you'd like me to do something, I'm prepared to dedicate some time for it this week (next weeks I'm away).

PR #923, opened 2024-10-04
Add check-node-configuration command

### LINKED ISSUE #922: Add check-node-configuration command
# Context

We want to add a `check-node-configuration` command to `cardano-cli` to validate the hashes of genesis files in the node's configuration file. This will remove some work from `cardano-testnet`, and at the same time providing support to users. This has been discussed on https://github.com/input-output-hk/cardano-node-wiki/pull/52.

# Design decisions

1. The command is in the `debug` top-level entry, because it is not dependent upon an era.
2. The command has an optional `--fix` flag that will correct the hashes if they are incorrect, or write them if they are missing. By default this option is OFF, because we want this command to be a checker first and foremost.
3. If a genesis hash is present, but the path to the genesis file is missing, the command doesn't fail, but issues a warning on stdout.
4. byron is handled, and this is a pain, because its handling is different, but :shrug:

smelc:
# Changelog

```yaml
- description: |
    Add the `debug check-node-configuration --node-configuration-file node-config.json` command
    that reads `node-config.json` and checks that the hashes of genesis files are correct.
  type:
```

# Context

* Fixes https://github.com/IntersectMBO/cardano-cli/issues/922
* Decision to write this command taken in https://github.com/input-output-hk/cardano-node-wiki/pull/52

# How to trust this PR

Try it yourself:

* Extract the [922.zip](https://github.com/user-attachments/files/17555119/922.zip) file
* Run the command on the zip's content: `cabal run cardano-cli -- debug check-node-configuration --node-configuration-file node-config.json`
* Try to change the hashes with the `node-config.json` file. Rerun the command and observe that it detects the change.

# Checklist

- [X] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,177 @@
...
+-- | The JSON key of the genesis hash for the given era.
+eraToGenesisHashKey :: AnyCardanoEra -> Text.Text
+eraToGenesisHashKey (AnyCardanoEra era) = eraToStringKey era <> "GenesisHash"
+
+-- | The JSON key of the genesis file for the given era.
+eraToGenesisFileKey :: AnyCardanoEra -> Text.Text
+eraToGenesisFileKey (AnyCardanoEra era) = eraToStringKey era <> "GenesisFile"
+
+-- | Part of the JSON keys associated with the given era.
+eraToStringKey :: CardanoEra a -> Text.Text
```
carbolymer: We already have this functionality implemented using
```haskell
docToText . pretty
```

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,177 @@
...
+      if looksLikeJson
+        then LBS.writeFile configFilePath $ Aeson.encodePretty nodeConfig'
+        else Yaml.encodeFile configFilePath nodeConfig'
+  liftIO $ putStrLn $ "Successfully checked node configuration file: " <> configFilePath
+  return ()
+ where
+  configFilePath = unFile configFile
+  eras :: [AnyCardanoEra] = [minBound .. maxBound]
+
+data Analysis = Analysis
```
carbolymer: This type is exported so I'd suggest using more distinctive and descriptive name for example `NodeConfigAnalysisResult`.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,177 @@
...
+  { concernedEra :: AnyCardanoEra
+  , genesisFilepath :: FilePath
+  -- ^ The path to the genesis file, as found in the node configuration file, if any
+  , actualHash :: Text.Text
+  -- ^ The actual hash of the genesis file, computed from the file's content
+  , expectedHash :: Maybe Text.Text
+  -- ^ The hash of the genesis file, as found in the node configuration file, if any.
+  }
+
+runAnalysis
```
carbolymer: I have the same remark as for `Analysis` type here.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: Changed :+1:

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,170 @@
...
+import qualified Data.Text.IO as Text
+import qualified Data.Yaml as Yaml
+import           System.Directory (removeFile)
+import           System.FilePath (takeDirectory, (</>))
+
+runCheckNodeConfig :: CheckNodeConfigCmdArgs -> ExceptT DebugCmdError IO ()
+runCheckNodeConfig (CheckNodeConfigCmdArgs configFile fixTheFile) = do
+  -- Use Cardano.CLI.EraBased.Run.Genesis.writeGenesisHashesToNodeConfigFile ?
+  -- and CreateTestnetData.addOrCheck?
+  nodeConfigValue <- Yaml.decodeFileThrow configFilePath
```
carbolymer: Why this type can't be used here https://github.com/IntersectMBO/cardano-api/blob/013cce9a1ebaee292572f8a35c789e8b0161add5/cardano-api/internal/Cardano/Api/LedgerState.hs#L1048 ?

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,170 @@
...
+      Aeson.Object obj -> pure obj
+      _ -> throwE $ DebugNodeConfigReadCmdError configFilePath "Expected an Object, but got something else"
+  analyses <- catMaybes <$> mapM (\era -> mkAnalysis era configFile nodeConfig) eras
+  nodeConfig' <- foldM (\obj -> runAnalysis configFile obj fixTheFile) nodeConfig analyses
+  when (nodeConfig' /= nodeConfig) $ do
+    -- Write the fixed node configuration file
+    -- Better to remove the file first, to avoid a file busy exception when writing
+    liftIO $ removeFile configFilePath
+    -- Cheap trick to try to preserve the input file's format
+    let looksLikeJson = ".json" `isSuffixOf` configFilePath
```
carbolymer: I think I saw `configuration.yaml` file in cardano-testnet, which contained json. I can't think of a better heuristic here though.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,170 @@
...
+    liftIO $
+      if looksLikeJson
+        then LBS.writeFile configFilePath $ Aeson.encodePretty nodeConfig'
+        else Yaml.encodeFile configFilePath nodeConfig'
+  liftIO $ putStrLn $ "Successfully checked node configuration file: " <> configFilePath
+  return ()
```
carbolymer:
```suggestion
```

carbolymer (commented):
LGTM. I think I'll leave the approval to @Jimbo4350

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: No it's not!

See https://github.com/IntersectMBO/cardano-cli/blob/smelc/check-node-config/cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs#L6:

```hs
module Cardano.CLI.Run.Debug.CheckNodeConfiguration (runCheckNodeConfig, eraToStringKey) where
```

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: See https://github.com/IntersectMBO/cardano-cli/pull/923#discussion_r1795316865

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: > I think I saw configuration.yaml file in cardano-testnet, which contained json

Yes that's the case, and I intend to change that soon :wink:

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: The problem with `LedgerState.NodeConfig` is that it supposes the hashes are present (except for Conway): 

image

So you can't leave them out and call `check-node-configuration --fix` to fill them in (which would be handy from `cardano-testnet`). Anyway, I'm fine either way (keep current code and ability to omit hashes; or improve code by using `NodeConfig`, but loose ability to fill in hashes).

@Jimbo4350> what do you think?

Jimbo4350: `Cardano.Api.LedgerState` has duplication in it wrt the node's configuration internals in `cardano-node` and we haven't decided yet how we want to deal with this.

What we really want is to mimic what is done in the node: https://github.com/IntersectMBO/cardano-node/blob/371a1427afc4ec4e19417c2e30baf1fe587b792e/cardano-node/src/Cardano/Node/Run.hs#L142

smelc: > What we really want is to mimic what is done in the node: IntersectMBO/cardano-node@371a142/cardano-node/src/Cardano/Node/Run.hs#L142

Yeah, in the long run we want that I know, but can we merge this PR soon? Moving `PartialNodeConfiguration` and friends is not going to happen quickly, so I need a decision on this PR's current state.

Jimbo4350: I think relying on `NodeConfig` is the right choice. Otherwise we introduce more logic to check the validity of a node configuration.

smelc: > I think relying on NodeConfig is the right choice

Sounds good, will change this PR to do that!

smelc:
Putting back to draft while I change the code to use `NodeConfig`.

smelc:
@Jimbo4350, @carbolymer> done that: https://github.com/IntersectMBO/cardano-cli/pull/923#discussion_r1795507988

However, the code still needs to use the vanilla `Aeson.Value` version of the node configuration, to write it back (when `--fix` is passed). We need that, because, as `NodeConfig` only covers a subset of the node's configuration file, we would only write back this subset when fixing the file.

Nevertheless, this version using `NodeConfig` for deserialization is way better than the initial one, because indeed we don't have to check the shape of what we read from disk.

cardano-cli/test/cardano-cli-test/Test/Cli/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,150 @@
...
+            , "test/cardano-cli-test/files/input/check-node-configuration/node-config.yaml"
+            ]
+        , fiddleKind <- [minBound .. maxBound]
+        ]
+
+  propertyOnce $ forM_ supplyValues $ \(nodeConfigPath, fiddleKind) -> H.moduleWorkspace "tmp" $ \tempDir ->
+    bracketSem nodeConfigSem $ \_ -> do
+      let finalInputConfig = tempDir </> "node-config-changed.json"
+          -- TODO why is that writing to "AlonzoGenesisHash" writes one more 0 than specified here? (and hence the need for this hack)
+          wrongHash = Text.pack $ replicate (case fiddleKind of FiddleByron -> 65; FiddleNonByron -> 64) '0'
```
smelc, line 71: This needs to be sorted out before merging. This one is actually quite surprising.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
Jimbo4350: I agree `Analysis` is ambiguous. Analysis of what?

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,188 @@
...
+  :: NodeConfigFile 'In
+  -- ^ The node configuration file. It's not read by this function, but used for producing error messages.
+  -> FixNodeConfigFile
+  -> Aeson.Object
+  -- ^ The node configuration file, as a JSON object. It will get written back to disk.
+  -> Analysis
+  -> ExceptT DebugCmdError IO Aeson.Object
+  -- ^ The fixed node configuration file
+runAnalysis configFile fixMode nodeConfigObj (Analysis{concernedEra = era@(AnyCardanoEra cEra), genesisFilepath = _, actualHash, expectedHash}) = do
+  case (expectedHash, fixMode) of
```
Jimbo4350: I think the fix mode is not worth it. We have all this additional clunky logic to insert into maps etc. Reporting the errors is enough and if the hash does not turn out to be the expected hash the user is probably going to take the time to investigate why vs just fixing it automatically.

cardano-cli/src/Cardano/CLI/Types/Errors/DebugCmdError.hs
```diff
@@ -7,13 +7,46 @@ where
...
   = DebugCmdFailed
+  | -- | @DebugNodeConfigGenesisDataCmdError filepath error@ represents an error when
+    -- reading the node configuration at @filepath@
+    DebugNodeConfigGenesisDataCmdError !FilePath !GenesisDataError
+  | -- | @DebugNodeConfigReadCmdError filepath error@ represents an error when
+    -- reading the node configuration at @filepath@
+    DebugNodeConfigReadCmdError !FilePath !String
+  | -- | @DebugNodeConfigWrongGenesisHashCmdError filepath era actualHash expectedHash@ represents a user error
+    -- that the genesis hash for @era@ in @filepath@ is @actualHash@, whereas it should be @expectedHash@
+    DebugNodeConfigWrongGenesisHashCmdError !FilePath AnyCardanoEra !Text !Text
```
Jimbo4350: Some haddocks to describe what `Text` values are would be useful.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,188 @@
...
+-- Reads the genesis file path and hash from the node configuration file, if they exist.
+mkAnalysis
+  :: AnyCardanoEra
+  -- ^ The era to analyze
+  -> NodeConfigFile 'In
+  -- ^ The node configuration file path. It's not read by this function, but used for producing error messages.
+  -> NodeConfig
+  -- ^ The parsed node configuration file
+  -> ExceptT DebugCmdError IO (Maybe Analysis)
+mkAnalysis cEra@(AnyCardanoEra era) configFile nodeConfig = do
```
Jimbo4350: We're doing a validation of the node configuration file. Why not use https://hackage.haskell.org/package/validation?

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,188 @@
...
+      liftIO $ Text.putStrLn $ "Replacing " <> expectedHash' <> " with " <> actualHash
+      pure $
+        Aeson.insert (Aeson.fromText $ eraToGenesisHashKey era) (Aeson.String actualHash) nodeConfigObj
+ where
+  configFilePath = unFile configFile
+
+-- | Build an 'Analysis' value from the given node configuration file, for the given era.
+-- Reads the genesis file path and hash from the node configuration file, if they exist.
+mkAnalysis
+  :: AnyCardanoEra
```
Jimbo4350: Use the experimental api interface here. We don't care about previous eras.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,188 @@
...
+
+data HashRole
+
+-- | A GADT, to be able to cover both files and hashes in a single function, e.g. 'getData' below
+data Role r where
+  FileRole :: Role FileRole
+  HashRole :: Role HashRole
+
+-- | A type family, to be able to cover both files and hashes in a single function, e.g. 'getData' below
+type family DataKind r where
```
Jimbo4350: This is overkill.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,188 @@
...
+  DataKind FileRole = FilePath
+  DataKind HashRole = Text.Text
+
+-- | Get the data associated with the given role for the given era.
+getData :: CardanoEra a -> Role r -> NodeConfig -> Maybe (DataKind r)
+getData era role config =
```
Jimbo4350: Create standalone functions that retrieve the genesis files and hashes from the node configuration. Pattern matching on the eras exposed by the experimental and calling the relevant functions will simplify this code.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: This parameter isn't about the era we support, it's about the era whose hash we want to validate. I updated the haddock to make this explicit, as this indeed is an usual use case for us.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: We don't depend on `validation` yet and I believe this simple use case does not justify adding an external dependency.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: Removed :+1: (it was the simplest use of a `type family` though :wink:)

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: As I wrote in https://github.com/IntersectMBO/cardano-cli/pull/923#discussion_r1800855128, this code doesn't depend on the era, so using the experimental API here doesn't help.

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
smelc: Removed the `fix` mode :+1:

cardano-cli/src/Cardano/CLI/Types/Errors/DebugCmdError.hs
smelc: Added haddocks to the individual arguments :+1:

cardano-cli/src/Cardano/CLI/Run/Debug/CheckNodeConfiguration.hs
Jimbo4350: We only care about mainnet and the upcoming era and therefore the configuration files associated with those eras which include Byron, Shelley, Alonzo and Conway genesis files/hashes. I.e we check all for the presence of all of theses things in the Conway era. If another genesis file is added in an upcoming era then we would add that additional check for the upcoming era. Currently there are a bunch of irrelevant cases that return `Nothing` which we can totally exclude by not parameterizing on `AnyCardanoEra`.

Jimbo4350 (changes requested):
https://github.com/IntersectMBO/cardano-cli/pull/923#discussion_r1801125921 

Also we need to come up with a better name than `Analysis`.

smelc:
> Also we need to come up with a better name than Analysis.

Renamed to `CheckGenesisAnalysis`, but please note that this name is not exported, so using a short name was fine IMO :slightly_smiling_face:

cardano-cli/test/cardano-cli-test/Test/Cli/CheckNodeConfiguration.hs
```diff
@@ -0,0 +1,127 @@
...
+                   newFileSem)
+
+import           Hedgehog (Property)
+import qualified Hedgehog as H
+import           Hedgehog.Extras (propertyOnce)
+import qualified Hedgehog.Extras as H
+
+-- | Semaphore protecting against locked file error, when running properties concurrently.
+nodeConfigSem :: FileSem
+nodeConfigSem = newFileSem "test/cardano-cli-test/files/input/check-node-configuration/node-config.json"
```
Jimbo4350, line 30: Are these tests actually run concurrently?

Jimbo4350 (approved):
LGTM! One requested change.

cardano-cli/test/cardano-cli-test/Test/Cli/CheckNodeConfiguration.hs
smelc: Yes parallelism happens even within a single file. cc @carbolymer

PR #937, opened 2024-10-15
Add hash checks for `governance committee create-cold-key-resignation-certificate` and `governance vote create`

### LINKED ISSUE #882: [FR] - Add File Hash Validation when Building Transaction

palas:
# Changelog

```yaml
- description: |
    Added hash checks for `governance committee create-cold-key-resignation-certificate` and `governance vote create`
  type:
  - feature
  - test
```

# Context

This is another step forward on addressing https://github.com/IntersectMBO/cardano-cli/issues/882.

This PR adds checks to two commands:
- `governance committee create-cold-key-resignation-certificate`
- `governance vote create`

It also generalizes the `carryHashChecks` function to be able to reuse it more.

# How to trust this PR

There are tests for all new functionality (both positive and negative). Most code is reused, and the one that is not is completely analogous to other PRs.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
@@ -17,6 +17,7 @@ import           Cardano.Api
 import qualified Cardano.Api.Ledger as L
 import           Cardano.Api.Shelley
 
+import           Cardano.CLI.Types.Common (PotentiallyCheckedAnchor, ResignationMetadataUrl)
 import           Cardano.CLI.Types.Key
 import           Cardano.CLI.Types.Key.VerificationKey
 
@@ -71,7 +72,13 @@ data GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs era
   = GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs
   { eon :: !(ConwayEraOnwards era)
   , vkeyColdKeySource :: !(VerificationKeySource CommitteeColdKey)
-  , anchor :: !(Maybe (L.Anchor (L.EraCrypto (ShelleyLedgerEra era))))
+  , anchor
+      :: !( Maybe
+              ( PotentiallyCheckedAnchor
+                  ResignationMetadataUrl
+                  (L.Anchor (L.EraCrypto (ShelleyLedgerEra era)))
+              )
+          )
   , outFile :: !(File () Out)
   }
   deriving Show
--- a/cardano-cli/src/Cardano/CLI/Types/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/Types/Common.hs
@@ -90,7 +90,6 @@ module Cardano.CLI.Types.Common
   , DRepMetadataUrl
   , ResignationMetadataUrl
   , PotentiallyCheckedAnchor (..)
-  , _pcaAnchor
   )
 where
 
@@ -105,7 +104,6 @@ import           Data.String (IsString)
 import           Data.Text (Text)
 import qualified Data.Text as Text
 import           Data.Word (Word64)
-import           Lens.Micro.Type (Traversal)
 
 -- | Determines the direction in which the MIR certificate will transfer ADA.
 data TransferDirection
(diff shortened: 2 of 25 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
```diff
@@ -71,7 +72,13 @@ data GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs era
   = GovernanceCommitteeCreateColdKeyResignationCertificateCmdArgs
   { eon :: !(ConwayEraOnwards era)
   , vkeyColdKeySource :: !(VerificationKeySource CommitteeColdKey)
-  , anchor :: !(Maybe (L.Anchor (L.EraCrypto (ShelleyLedgerEra era))))
+  , anchor
+      :: !( Maybe
```
Jimbo4350, line 76: I'm really not a fan of  `PotentiallyCheckedAnchor` because this is an input argument submitted by the user and _they are deciding_ whether or not to check it.  IIRC @smelc also was leaning towards having two separate arguments for this. We don't need to deal with this in this PR but the name is definitely misleading. Having a resulting type that indicates if the hash has been checked or not would be more appropriate vs `PotentiallyCheckedAnchor` which is telling us if we still have to check the hash or not. A user could technically submit an anchor and trust it without having actually checked it themselves.

cardano-cli/src/Cardano/CLI/Types/Common.hs
```diff
@@ -665,3 +667,12 @@ data PotentiallyCheckedAnchor anchorType anchor
   -- ^ Whether to check the hash or not (CheckHash for checking or TrustHash for not checking)
   }
   deriving (Eq, Show)
+
+_pcaAnchor
```
Jimbo4350: Please remove this. We don't need lenses.

Jimbo4350 (changes requested):
Please remove `_pcaAnchor` as it's not strictly necessary. I want our lens usage constrained to what the ledger api exposes.

cardano-cli/src/Cardano/CLI/EraBased/Commands/Governance/Committee.hs
palas: Do you mean the name of the parameter the user writes should be called differently? Or that the code should be structured different? The reason the type is wrapped in `PotentiallyCheckedAnchor` is that the option to check the hash only makes sense if there is a hash, and it is optional, so this way we can disallow providing the check requirement and not the hash

cardano-cli/src/Cardano/CLI/Types/Common.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/937/commits/525b7d325367a896e89913dea634e5d2e8e38100

cardano-cli/test/cardano-cli-golden/Test/Golden/Governance/Vote.hs
```diff
@@ -179,3 +185,54 @@ hprop_golden_governance_governance_vote_create_abstain_cc_hot_key =
...
+              , voteFile
+              , "--anchor-url"
+              , "ipfs://" ++ exampleAnchorDataIpfsHash
+              , "--anchor-data-hash"
+              , alteredHash
+              , "--check-anchor-data-hash"
+              ]
+        )
+
+    exitCode === ExitFailure 1
```
Jimbo4350, line 236: In theory this can fail for other reasons (drep verification key does not exist, misspelled cli arguments etc). We should test the output of the failure. Not in this PR but in a follow up PR. We can do something simple like confirming `Hashes do not match!` exists in the output.

Jimbo4350 (approved):

cardano-cli/test/cardano-cli-golden/Test/Golden/Governance/Vote.hs
palas: Ah, but that is done underneath, cause it is a golden test

palas: `H.diffVsGoldenFile result voteGold`

PR #951, opened 2024-10-22
Add anchor hash checks to `transaction build`

### LINKED ISSUE #882: [FR] - Add File Hash Validation when Building Transaction

palas:
# Changelog
```yaml
- description: |
    Added anchor data hash checks to `transaction build`
  type:
  - feature
```

# Context

This is part of work related to solving https://github.com/IntersectMBO/cardano-cli/issues/882. Each individual command has optional checks for verifying the hash matches the anchor data. But this PR adds a check to `transaction build` which combines the pieces and creates the actual transaction (so it is one of the last steps).

## Related PRs

- [cardano-api] https://github.com/IntersectMBO/cardano-api/pull/664
- [cardano-node] https://github.com/IntersectMBO/cardano-node/pull/6023

# How to trust this PR

You can see the new test in [this PR in `cardano-testnet`](https://github.com/IntersectMBO/cardano-node/pull/6023) and the old tests updated to pass.

# Checklist

- [x] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
carbolymer, line 1: 1. This module is in the wrong place. In `EraBased` module we should only have `Commands`, `Options` and `Run`. Since this is used by a transaction only, I think it should go to `Cardano.CLI.EraBased.Transaction.HashCheck`
2. A base form instead of participle should be used for module name, meaning: `HashCheck` instead of `HashChecking`

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
```diff
@@ -0,0 +1,123 @@
...
+          { pcaMustCheck = CheckHash
+          , pcaAnchor = anchor
+          }
+      )
+
+-- | Find references to anchor data and check the hashes are valid
+-- and they match the linked data.
+checkCertificateHashes :: Certificate era -> ExceptT TxCmdError IO ()
+checkCertificateHashes c =
+  case c of
```
carbolymer: The code looks correct by itself, but this way of access of certificates is problematic. We should be using ledger patterns when accessing certificates. See how it's implemented here: https://github.com/IntersectMBO/cardano-api/blob/main/cardano-api/internal/Cardano/Api/Certificate.hs#L518

This function should be moved to `Cardano.Api.Certificate` in order to have handling of certificates in one place. We had multiple issues related to small mistakes related to ledger certificates, so considering future maintenance, it's better to keep it in one place.

cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
```diff
@@ -217,6 +222,10 @@ renderTxCmdError = \case
     prettyError e
   TxCmdFeeEstimationError e ->
     prettyError e
+  TxCmdPoolMetadataHashError url ->
+    "Hash of the pool metadata file is not valid. Url:" <+> pretty (L.urlToText url)
+  TxCmdHashCheckError url e ->
+    "Hash of the file is not valid. Url:" <+> pretty (L.urlToText url) <+> pretty (displayException e)
```
carbolymer:
```suggestion
    "Hash of the file is not valid. Url:" <+> pretty (L.urlToText url) <+> prettyException e
```

carbolymer (changes requested):

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
```diff
@@ -0,0 +1,123 @@
...
+              L.ConwayUnRegDRep _ _ -> return ()
+              L.ConwayUpdateDRep _ mAnchor -> forM_ mAnchor checkAnchorMetadataHash
+              L.ConwayAuthCommitteeHotKey _ _ -> return ()
+              L.ConwayResignCommitteeColdKey _ mAnchor -> forM_ mAnchor checkAnchorMetadataHash
+ where
+  checkPoolMetadataHash :: L.PoolMetadata -> ExceptT TxCmdError IO ()
+  checkPoolMetadataHash (L.PoolMetadata{L.pmUrl = url, L.pmHash = hashBytes}) = do
+    let mHash = L.hashFromBytes hashBytes
+    hash <- maybe (left $ TxCmdPoolMetadataHashError url) return mHash
+    let safeHash = L.unsafeMakeSafeHash hash
```
carbolymer: >-- | Don't use this except in Testing to make Arbitrary instances, etc.
>--   Defined here, only because the Constructor is in scope here.

This comment on `L.unsafeMakeSafeHash` is a bit worrying. But I see that ledger does not expose a nice way of getting the pool hash.

Maybe we should provide an instance:
```
instance HashWithCrypto PoolMetadata PoolMetadata
```
to be able to calculate the hash the right way? I'm guessing here.

This hash calculation logic should be implemented as a function in `cardano-api` I think. Maybe even upstreamed to ledger?

carbolymer (commented):
I think those hash checking bits should go to `cardano-api`. It feels like it's a reusable code, which we could expose to api users.

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
palas: Yes, I think `SafeHash` is widely misused (doesn't really have any guarantees anymore). The fact that there are parsers that produce `SafeHash` makes no sense to me, because there is no way they check the hash corresponds to something. But some functions do require `SafeHash` for some reason, when they shouldn't, so there is no way around it.

palas: The idea is `SafeHash` must internally store the data that hashes to that (as I understand it), to ensure it is a hash of something. But here we are reading the actual hash (not the thing that is hashed) from command line, so we cannot get that info. But we do need a `SafeHash` 🤷‍♂️

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
```diff
@@ -0,0 +1,123 @@
...
+    Shelley.shelleyBasedEraConstraints eon $ do
+      checkAnchorMetadataHash anchor
+      checkGovActionHashes govAction
+   where
+    checkGovActionHashes
+      :: L.GovAction (Shelley.ShelleyLedgerEra era) -> ExceptT TxCmdError IO ()
+    checkGovActionHashes govAction' =
+      Shelley.shelleyBasedEraConstraints eon $
+        case govAction' of
+          L.ParameterChange{} -> return ()
```
Jimbo4350: A comment stating that only the `NewConstitution` governance action contains a checkable hash would be useful.

Jimbo4350 (commented):
I agree with @carbolymer . Those changes need to be made first.

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
Jimbo4350: Just discussed with @palas that we should be not unsafely constructing `SafeHash`s. I can't envision a specific scenario for this but I want to play it safe ( :wink: :sunglasses: ) and eliminate the problem of potentially  depending on the original bytes from an unsafely constructed `SafeHash`. @palas can you link to the issue after you've created it?

cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
palas: Nice, addressed in: https://github.com/IntersectMBO/cardano-cli/commit/5784d620dc43415444d8c0f3b212c5c97c063ab6

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/951/commits/5784d620dc43415444d8c0f3b212c5c97c063ab6
and: https://github.com/IntersectMBO/cardano-api/pull/664

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/commit/5784d620dc43415444d8c0f3b212c5c97c063ab6

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/951/commits/78f05e0d79706772760b447d976ff27b59f47763

cardano-cli/src/Cardano/CLI/EraBased/HashChecking.hs
palas: Issue added here: https://github.com/IntersectMBO/cardano-cli/issues/952

Jimbo4350 (approved):
LGTM but I don't agree with `AnchorDataFromCertificateException` as a name. I made a comment on the relevant api PR.

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+
+module Cardano.CLI.EraBased.Transaction.HashCheck
+  ( checkCertificateHashes
+  , checkVotingProcedureHashes
+  , checkProposalHashes
+  )
+where
+
+import           Cardano.Api (Certificate (..), ExceptT, firstExceptT)
+import qualified Cardano.Api as L
```
carbolymer:
```suggestion
import Cardano.Api
```
I think this shouldn't be imported qualified `as L`

cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
```diff
@@ -217,6 +221,10 @@ renderTxCmdError = \case
     prettyError e
   TxCmdFeeEstimationError e ->
     prettyError e
+  TxCmdPoolMetadataHashError e ->
+    "Hash of the pool metadata hash is not valid:" <+> prettyException e
+  TxCmdHashCheckError url e ->
+    "Hash of the file is not valid. Url:" <+> pretty (L.urlToText url) <+> prettyException e
```
carbolymer, line 227: `instance Pretty L.Url` could be added to `cardano-api`.

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+-- | Find references to anchor data in voting procedures and check the hashes are valid
+-- and they match the linked data.
+checkVotingProcedureHashes
+  :: Shelley.ShelleyBasedEra era -> Shelley.VotingProcedures era -> ExceptT TxCmdError IO ()
+checkVotingProcedureHashes eon (Shelley.VotingProcedures (L.VotingProcedures voterMap)) =
+  Shelley.shelleyBasedEraConstraints eon $
+    forM_
+      voterMap
+      ( \vpMap ->
+          forM_
```
carbolymer, line 48: I think `mapM_` could be more readable here

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+            vpMap
+            ( \(L.VotingProcedure _ mAnchor) ->
+                forM_ mAnchor checkAnchorMetadataHash
+            )
+      )
```
carbolymer:
```suggestion
      ( mapM $\(L.VotingProcedure _ mAnchor) ->
                forM_ mAnchor checkAnchorMetadataHash
      )
```
I think mapM may be more readable here

carbolymer (approved):

cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
palas: Addressed in rebase

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
palas: Addressed in rebase

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
palas: Applied in commit: https://github.com/IntersectMBO/cardano-cli/pull/951/commits/d854349759700ad18089ae84f6157e1ba9f0b8b2

cardano-cli/src/Cardano/CLI/EraBased/Transaction/HashCheck.hs
palas: Addressed in rebase

cardano-cli/src/Cardano/CLI/Types/Errors/TxCmdError.hs
palas: [Ticket](https://input-output-rnd.slack.com/lists/T0N639Z4N/F078RTQGB1T?record_id=Rec07U4MFSTU6)

gitmachtl:
@palas @carbolymer can we please have a flag that disables this automated anchor hash check?

we just had a breaking change that the cli requires a new env variable set for the ipfs gateway uri. this was not mentioned anywhere in the changelogs. 

also, some might not want the the cli is reaching out to the internet on its own. so can we please have a flag to disable automated online-checks for http/https/ipfs anchors? can still be enabled by default imo.

PR #968, opened 2024-11-14
create-testnet-data: use experimental API and make arguments era specific

### LINKED ISSUE #962: More fine grained era control in create-testnet-data
I would parameterize it on `Era era`. We actually need to deprecate the `BabbageEra` constructor of `Era era`.

_Originally posted by @Jimbo4350 in https://github.com/IntersectMBO/cardano-cli/pull/961#discussion_r1831298543_

smelc:
# Changelog

```yaml
- description: |
    More fine grained controls of eras for create-testnet-data. Forbid creating DReps and CC in babbage.

    Remove the `create-testnet-data` altogether in eras earlier than Babbage
  type:
```

# Context

Fixes https://github.com/IntersectMBO/cardano-cli/issues/962

# How to trust this PR

Look at the changes in golden files

# Checklist

- [X] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Commands/Genesis.hs
```diff
@@ -127,7 +131,9 @@ data GenesisCreateTestNetDataCmdArgs era = GenesisCreateTestNetDataCmdArgs
   , outputDir :: !FilePath
   -- ^ Directory where to write credentials and files.
   }
-  deriving Show
+
+instance Show GenesisCreateTestNetDataCmdArgs where
+  show _ = "GenesisCreateTestNetDataCmdArgs"
```
palas, line 136: It doesn't derive anymore? It is odd because `Era` does have a `Show` instance ([here](https://github.com/IntersectMBO/cardano-api/blob/76fba1ca226e3cd48e7d42f947da0d9aa1b8f117/cardano-api/internal/Cardano/Api/Experimental/Eras.hs#L98)), maybe after moving the parameter it will?

cardano-cli/src/Cardano/CLI/EraBased/Options/Genesis.hs
```diff
@@ -217,18 +218,18 @@ pGenesisCreateStaked sbe envCli =
...
       <$> optional (pSpecFile "shelley")
       <*> optional (pSpecFile "alonzo")
       <*> optional (pSpecFile "conway")
       <*> pNumGenesisKeys
       <*> pNumPools
       <*> pNumStakeDelegs
-      <*> pNumCommittee
-      <*> pNumDReps
+      <*> (case era of Exp.BabbageEra -> pure 0; Exp.ConwayEra -> pNumCommittee) -- Committee doesn't exist in babbage
+      <*> (case era of Exp.BabbageEra -> pure $ DRepCredentials OnDisk 0; Exp.ConwayEra -> pNumDReps) -- DReps don't exist in babbage
```
palas, line 232: Making it unrepresentable on the type would be safer. But this is probably much simpler, so it seems a reasonable trade-off

palas (approved):
Looks good. I left a couple of comments on things that caught my eye, just minor things. Feel free to address them if you think they make sense, but I leave it to your judgment

palas:
Ah, one more thing, probably say what got broken in the change log. (That some eras no longer have the command.)

cardano-cli/src/Cardano/CLI/EraBased/Commands/Genesis.hs
smelc: @palas> it doesn't derive anymore because there's an existential type in `GenesisCreateTestnetDataCmdArgs` now: 

image

cardano-cli/src/Cardano/CLI/EraBased/Commands/Genesis.hs
```diff
@@ -92,8 +94,10 @@ data GenesisCreateStakedCmdArgs era = GenesisCreateStakedCmdArgs
   }
   deriving Show
 
-data GenesisCreateTestNetDataCmdArgs era = GenesisCreateTestNetDataCmdArgs
-  { eon :: !(ShelleyBasedEra era)
+-- TODO This existential type parameter should become a regular type parameter
+-- when we parameterize the parent type by the experimental era API.
+data GenesisCreateTestNetDataCmdArgs = forall era. GenesisCreateTestNetDataCmdArgs
+  { eon :: !(Exp.Era era)
```
carbolymer, line 100: Why not `Some Exp.Era` instead?

cardano-cli/src/Cardano/CLI/EraBased/Options/Genesis.hs
carbolymer: Those inlined cases are not very readable, can you move them to a values in `where`, and just reference them from here?

carbolymer: I'd also use `Ord (Some Era)` instead of pattern matching on each era. I mean:
```haskell
if someEra < Some Exp.ConwayEra 
  then pure 0 
  else pNumCommittee
```
and
```haskell
if someEra < Some Exp.ConwayEra
  then pure $ DRepCredentials OnDisk 0
  else pNumDReps
```
Rationale: less maintenance burden when introducing new eras - you won't have to change each usage site of `case`.

smelc: > I'd also use Ord (Some Era) instead of pattern matching on each era. I mean:

@carbolymer> I'm not sure this is really better, because there are at most two _experimental_ eras at any point in time and when we delete one era, the index of the other era (the one being kept) shrinks from 1 to 0. So I wouldn't rely on the ordering here. I think pattern matching makes the intent more explicit.

PR #979, opened 2024-11-26
Allow and add warning for HTTP anchor-data

palas:
# Changelog

```yaml
- description: |
    Modified anchor-data checking to allow HTTP schema for testing purposes
  type:
  - feature
```

# Context

Having only support for HTTPS and IPFS is cumbersome, because it would require obtaining an SSL/TLS certificate and a domain for running the tests, or publishing to IPFS with every test (whenever the anchor-data changes).

This PR allows HTTP schema but shows a warning when it is used.

It also ensures we use "scheme" throughout and not "schema" (for self-consistency and consistency with IANA nomenclature), and adds documentations for some functions.

# How to trust this PR

There are plenty of tests so, in my opinion, it is mainly about documentation and code structure. I tested the warning manually, let me know if we should add an automated test for it.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/DRep.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/DRep.hs
@@ -20,7 +20,7 @@ import qualified Cardano.Api.Ledger as L
 
 import qualified Cardano.CLI.Commands.Hash as Cmd
 import qualified Cardano.CLI.EraBased.Commands.Governance.DRep as Cmd
-import           Cardano.CLI.Run.Hash (allSchemas, carryHashChecks, getByteStringFromURL)
+import           Cardano.CLI.Run.Hash (allSchemes, carryHashChecks, getByteStringFromURL)
 import qualified Cardano.CLI.Run.Key as Key
 import           Cardano.CLI.Types.Common
 import           Cardano.CLI.Types.Errors.CmdError
@@ -187,7 +187,7 @@ runGovernanceDRepMetadataHashCmd
       Cmd.DrepMetadataFileIn metadataFile ->
         firstExceptT ReadFileError . newExceptT $ readByteStringFile metadataFile
       Cmd.DrepMetadataURL urlText ->
-        fetchURLToGovernanceCmdError $ getByteStringFromURL allSchemas $ L.urlToText urlText
+        fetchURLToGovernanceCmdError $ getByteStringFromURL allSchemes $ L.urlToText urlText
     let (_metadata, metadataHash) = hashDRepMetadata metadataBytes
     case hashGoal of
       Cmd.CheckHash expectedHash
--- a/cardano-cli/src/Cardano/CLI/Run/Hash.hs
+++ b/cardano-cli/src/Cardano/CLI/Run/Hash.hs
@@ -9,10 +9,9 @@
 module Cardano.CLI.Run.Hash
   ( runHashCmds
   , getByteStringFromURL
-  , SupportedSchemas (..)
-  , allSchemas
-  , httpsAndIpfsSchemas
   , carryHashChecks
+  , allSchemes
+  , httpsAndIpfsSchemes
   )
 where
 
@@ -20,10 +19,13 @@ import           Cardano.Api
 import qualified Cardano.Api.Ledger as L
 
 import qualified Cardano.CLI.Commands.Hash as Cmd
+import           Cardano.CLI.Parser (stringToAnchorScheme)
 import           Cardano.CLI.Read
-import           Cardano.CLI.Types.Common (MustCheckHash (..), PotentiallyCheckedAnchor (..))
+import           Cardano.CLI.Types.Common (AnchorScheme (..), MustCheckHash (..),
+                   PotentiallyCheckedAnchor (..), SupportedSchemes)
 import           Cardano.CLI.Types.Errors.HashCmdError
 import           Cardano.Crypto.Hash (hashToTextAsHex)
+import           Cardano.Prelude (first)
 
 import           Control.Exception (throw)
 import           Control.Monad (when)
@@ -32,7 +34,6 @@ import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BS8
 import qualified Data.ByteString.Lazy as BSL
 import qualified Data.ByteString.Lazy.Char8 as BSL8
-import           Data.Char (toLower)
 import           Data.Function
 import           Data.List (intercalate)
 import           Data.Text (Text)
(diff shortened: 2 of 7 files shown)


cardano-cli/src/Cardano/CLI/Run/Hash.hs
```diff
@@ -94,30 +94,79 @@ runHashAnchorDataCmd Cmd.HashAnchorDataCmdArgs{toHash, hashGoal} = do
...
   fetchURLToHashCmdError = withExceptT HashFetchURLError
 
-data SupportedSchemas = FileSchema | HttpSchema | HttpsSchema | IpfsSchema
-  deriving (Show, Eq)
-
-allSchemas :: [SupportedSchemas]
-allSchemas = [FileSchema, HttpSchema, HttpsSchema, IpfsSchema]
+-- | Specifies the treatment for the different schemes when checking anchor data from URLs.
+data SupportedSchemes = SupportedSchemes
+  { supportedSchemes :: [AnchorScheme]
```
Jimbo4350: We are unnecessarily complicating this data definition. Adding `HttpScheme` to  `AnchorScheme` with a comment is fine.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Run/Hash.hs
palas: Addressed now, and squashed, because there was a bug in the previous commit

cardano-cli/src/Cardano/CLI/Run/Hash.hs
```diff
@@ -94,30 +94,56 @@ runHashAnchorDataCmd Cmd.HashAnchorDataCmdArgs{toHash, hashGoal} = do
...
-      | FileSchema `elem` supportedSchemas ->
-          let path = uriPathToFilePath (pathSegments uri)
-           in handleIOExceptT (FetchURLReadFileError path) $ BS.readFile path
-    "http:" | HttpSchema `elem` supportedSchemas -> getFileFromHttp uri
-    "https:" | HttpsSchema `elem` supportedSchemas -> getFileFromHttp uri
-    "ipfs:" | IpfsSchema `elem` supportedSchemas -> do
+  uri@URI{uriScheme} <- hoistMaybe (FetchURLInvalidURLError urlString) $ parseAbsoluteURI urlString
+  scheme <-
+    filterMaybe (`elem` supportedSchemes) (stringToScheme $ map toLower uriScheme)
+      ?? FetchURLUnsupportedURLSchemeError uriScheme
```
Jimbo4350: https://hackage.haskell.org/package/transformers-except-0.1.4/docs/Control-Monad-Trans-Except-Extra.html#v:hoistMaybe

cardano-cli/src/Cardano/CLI/Run/Hash.hs
```diff
@@ -94,30 +94,56 @@ runHashAnchorDataCmd Cmd.HashAnchorDataCmdArgs{toHash, hashGoal} = do
...
+-- not notice that the anchor-data being verified in the browser is not the same.
+httpsAndIpfsSchemes :: SupportedSchemes
+httpsAndIpfsSchemes =
+  [ HttpScheme -- Insecure, only for testing purposes
+  , HttpsScheme
+  , IpfsScheme
+  ]
+
+-- | Converts a string to an 'AnchorScheme' if it is a valid scheme, otherwise returns 'Nothing'.
+stringToScheme :: String -> Maybe AnchorScheme
```
Jimbo4350: This should really be moved to the cli parsers.

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/DRep.hs
```diff
@@ -187,7 +187,7 @@ runGovernanceDRepMetadataHashCmd
       Cmd.DrepMetadataFileIn metadataFile ->
         firstExceptT ReadFileError . newExceptT $ readByteStringFile metadataFile
       Cmd.DrepMetadataURL urlText ->
-        fetchURLToGovernanceCmdError $ getByteStringFromURL allSchemas $ L.urlToText urlText
+        fetchURLToGovernanceCmdError $ getByteStringFromURL allSchemes $ L.urlToText urlText
```
Jimbo4350, line 190: Why are we calling `urlToText` here?

cardano-cli/src/Cardano/CLI/Run/Hash.hs
```diff
@@ -94,30 +94,56 @@ runHashAnchorDataCmd Cmd.HashAnchorDataCmdArgs{toHash, hashGoal} = do
...
+stringToScheme "file:" = Just FileScheme
+stringToScheme "http:" = Just HttpScheme
+stringToScheme "https:" = Just HttpsScheme
+stringToScheme "ipfs:" = Just IpfsScheme
+stringToScheme _ = Nothing
+
+-- | Fetches the content of a URL as a 'ByteString'.
+-- The URL must be an absolute URL. The supported schemes are specified in the 'SupportedSchemes' argument.
+-- If the scheme is not supported, an error is thrown.
+getByteStringFromURL :: SupportedSchemes -> Text -> ExceptT FetchURLError IO BS.ByteString
```
Jimbo4350, line 117: Why is this a `Text` value? Why not have the cli parsers parse a `URL` along with it's scheme? We can fail there if necessary _at the point of parsing_. Then we shouldn't need to hardcode `httpsAndIpfsSchemes` and `allSchemes` and pass them around.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/EraBased/Run/Governance/DRep.hs
palas: This [`urlToText`](https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-core/Cardano-Ledger-BaseTypes.html#v:urlToText) is for unpacking this kind of [`URL`](https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-core/Cardano-Ledger-BaseTypes.html#v:urlToText), which really just packs a `Text` (you can see [here](https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-core/src/Cardano.Ledger.BaseTypes.html#Url)). So it doesn't do any validation, the one I used for validating in `getByteStringFromURL` is this one: [`URI`](https://hackage.haskell.org/package/network-uri-2.6.4.2/docs/Network-URI.html#t:URI). So I have to unwrap and re-wrap

cardano-cli/src/Cardano/CLI/Run/Hash.hs
palas: It is a good point. We are not always getting it from command line, in the case of `transaction build` we are extracting it from a `Ledger` type, which is already "parsed" as a wrapped `Text` (see [previous comment](https://github.com/IntersectMBO/cardano-cli/pull/979#discussion_r1858777662)) and we convert it to the `URI`. So it may be nicer to use just it if we get a `Text` as a parameter. But we could separate the parser and add a converter function and use those before calling this. In any case, I would like to do that as a separate PR if this PR is urgent

cardano-cli/src/Cardano/CLI/Run/Hash.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/979/commits/fc7e39146f535d003704d542af32764001cbf613

Jimbo4350 (approved):
LGTM but please remove the `errors` dependency. You can use `hoistMaybe`.

cardano-cli/src/Cardano/CLI/Run/Hash.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/979/commits/dd3518e46d5f006ac4561aa2d7856055270b6205

PR #982, opened 2024-11-28
Move `genesis hash` to `hash genesis-file`

palas:
# Changelog

```yaml
- description: |
    Moved `genesis hash` command to `hash genesis-file`
  type:
  - breaking
```

# Context

This command was only accessible under era specific sub-commands, but itself is era independent. So this PR moves it to the `hash` subcommand, which is era independent.

# How to trust this PR

Changes to golden files are most significative. The most important thing is probably the code remains the same in the new location, but it was copied. Then make sure the new command name is appropriate, and that you are happy with the old command locations not being available anymore.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/test/cardano-cli-golden/files/golden/help.cli
+++ b/cardano-cli/test/cardano-cli-golden/files/golden/help.cli
@@ -544,7 +544,6 @@ Usage: cardano-cli legacy genesis
                                     | create-cardano
                                     | create
                                     | create-staked
-                                    | hash
                                     )
 
   Genesis block commands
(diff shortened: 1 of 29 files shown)


smelc (approved):
Code LGTM so I'm approving but I'm wondering if we should leave the old command and just have it output `era genesis hash is deprecated, please use hash genesis-file` to be nice to users. And then only after the next release we remove it altogether. This is how we proceeded in the past.

cc @CarlosLopezDeLara

cardano-cli/test/cardano-cli-golden/files/golden/help.cli
```diff
@@ -544,7 +544,6 @@ Usage: cardano-cli legacy genesis
                                     | create-cardano
                                     | create
                                     | create-staked
-                                    | hash
```
Jimbo4350: Don't break this. Let's follow our deprecation plan. Print a deprecation warning and after two months we can remove it. It would be nice if CI could warn us but that's a nice to have and not required in this PR.

palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/982/commits/e5b9136d2b2b09452059ac494f4448272cfc311d

Issue #911, opened 2024-09-24
[FEATURE] -  Update `cardano-cli conway query spo-stake-distribution`


CarlosLopezDeLara:
## What
We need to update the output of  `cardano-cli conway query spo-stake-distribution` to include the DREP delegation choices of the Pool's reward account. This, once https://github.com/IntersectMBO/cardano-ledger/issues/4645 is implemented, in particular: 

```
To be more specific the solution to this problem is:

1.  To change the default vote for SPOs to `No` from `Abstain` for all of the three proposals in question
2. Check whether SPO's reward address is delegated to the predefined `AlwaysNoConfidence` DRep and count default vote as `Yes` only on `NoConfidence` proposals for that SPO.
3. Check whether SPO's reward address is delegated to predefined `AlwaysAbstain` DRep and count default vote as `Abstain` for that SPO on all three: `NoConfidence`, `UpdateCommitee` and `ParameterUpdate` proposals.
```
For example:

```json
    [
        "dec39dd553180bf78b206258aba3861537b185580086ca3f62716f8d",
        10868846712397,
        "drep-alwaysAbstain"
    ],
    [
        "e2fd1cd89bba55dc457574adf420240c28a2bfdb3d460279e453c57a",
        10671709098807,
        "drep-keyHash-ffc25bcec3379aea4e32f353b4048fc7ab9fc1124c01738546e2f7bb"
    ],
    [
        "eec73ca2652e1ae92852195ff3289dd991a83dbe5f989f06e1c96b64",
        10703582488560,
        "drep-alwaysNoConfidence"
    ],
    [
        "f386ee192cf53c26da7457b7d8992ae0e52a05fc1c88c5d653a90021",
        10466189795103,
        "drep-scriptHash-186e32faa80a26810392fda6d559c7ed4721a65ce1c9d4ef3e1c87b4"
    ]
```
(Thank you @gitmachtl) 

## Why

https://github.com/IntersectMBO/cardano-ledger/issues/4645 changes the way SPO votes are counted. The SPO reward account delegation choice will impact how the pool's stake counts for voting. It is important to show this information so that downstream tools can recreate the vote count. 

## Personas

- [x] SPOs
- [ ] dApp Devs
- [ ] Exchanges
- [ ] Wallets
- [ ] 3rd party tools
- [x] ADA holders 

## Definition of Done (DoD)

[...]

PR #987, opened 2024-12-04
create-testnet-data: fix generation of Conway genesis w.r.t. delegation of voting power

smelc:
# Changelog

```yaml
- description: |
    Fix that the `initialDReps` fields in the `conway-genesis.json` file generated by `create-testnet-data` was incomplete.
  type:
```

# Context

While testing https://github.com/IntersectMBO/cardano-cli/issues/911, I noticed that the SPO created by `create-testnet-data` doesn't delegate its voting power to a DRep.

So I looked at `create-testnet-data` and found that only the `--stake-delegators` created by `create-testnet-data` delegate to DReps (whereas the SPO is created by `--pools`). But then, starting with `--stake-delegators`, there is a bug (introduced in https://github.com/IntersectMBO/cardano-cli/pull/940/files#diff-46f6afb73150af450b10d232ebdca10061efd5a69e268f5b0870439aca3b19e0) that causes an incorrect `conway-genesis` to be created.

This PR fixes that

# How to trust this PR

TBD

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/cardano-cli.cabal
+++ b/cardano-cli/cardano-cli.cabal
@@ -371,6 +371,7 @@ test-suite cardano-cli-golden
   type: exitcode-stdio-1.0
   build-depends:
     aeson >=1.5.6.0,
+    aeson-pretty,
     base16-bytestring,
     bytestring,
     cardano-api:{cardano-api, gen},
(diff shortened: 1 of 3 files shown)


smelc:
This PR is useless, by virtue of the explanation here: https://github.com/IntersectMBO/cardano-ledger/issues/4782#issuecomment-2523609735

This PR was causing the following change: 

image

But it is not required, because, as @lehins explained; the field `delegators` is derived from `delegs`; so we don't need to populate it. I'll do another PR to add some comments to the code, so that we keep a trace of this, because this is confusing IMO.

PR #993, opened 2024-12-12
Query.hs: simplify implementation with a few new functions and code sharing

smelc:
# Changelog

```yaml
- description: |
    Simplify implementation of queries with a few new functions and code sharing
  type:
```

# Context

Code in `Query.hs` is hard to maintain and shuffle because of the noise of error handling. This PR aims to improve a bit on that.

# How to trust this PR

* It's refactoring only
* It's +70/-167 :slightly_smiling_face: 

# Checklist

- [X] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
```diff
@@ -1281,16 +1228,13 @@ runQueryStakePoolsCmd
...
+            AnyCardanoEra era <- easyRunQueryCurrentEra
 
             sbe <-
               requireShelleyBasedEra era
                 & onNothing (left QueryCmdByronEra)
 
-            poolIds <-
-              lift (queryStakePools sbe)
-                & onLeft (left . QueryCmdUnsupportedNtcVersion)
-                & onLeft (left . QueryCmdEraMismatch)
```
palas, line 1293: This instance doesn't seem equivalent because it doesn't seem to handle the same errors. Before the function extraction it handled `QueryCmdEraMismatch`, and `easyRunQuery` doesn't (it handles `QueryCmdLocalStateQueryError . EraMismatchError`).

palas (changes requested):
I think I found a potential issue (maybe it is fine, but worth double checking). Other than that it looks good

cardano-cli/src/Cardano/CLI/EraBased/Run/Query.hs
smelc: @palas> the change was correct because actually `QueryCmdLocalStateQueryError . EraMismatchError` was equivalent to `QueryCmdEraMismatch`. But your comment made me realize that the `QueryCmdEraMismatch` approach was simpler, so I added a commit deleting `QueryCmdLocalStateQueryError` altogether, because it had only one use (that we can replace with `QueryCmdEraMismatch`). Can you have a new look?

palas (approved):
Looks good 👍

PR #1005, opened 2025-01-06
transaction id: add --output-[json,text] flag to control format of the output

### LINKED ISSUE #1001: era transaction txid: add --output-json and --output-text
This is part of https://github.com/IntersectMBO/cardano-cli/issues/566

_Originally posted by @smelc in https://github.com/IntersectMBO/cardano-cli/issues/566#issuecomment-2504061238_

smelc:
# Changelog

```yaml
- description: |
    transaction id: add --output-[json,text] flag to control format of the output
  type:
```

# Context

* Part of fixing https://github.com/IntersectMBO/cardano-cli/issues/566
* Fixes https://github.com/IntersectMBO/cardano-cli/issues/1001
* Way to get me back to coding after the Christmas break :blush: 

# How to trust this PR

* New tests are added
* You can try the new test before the commit introducing the feature, and see it still works on `HEAD` so the default behavior didn't change

# Checklist

- [X] Self-reviewed the diff


cardano-cli/test/cardano-cli-golden/files/golden/help/allegra_transaction_txid.cli
```diff
@@ -2,10 +2,15 @@ Usage: cardano-cli allegra transaction txid
...
 
   Print a transaction identifier.
 
 Available options:
   --tx-body-file FILEPATH  Input filepath of the JSON TxBody.
   --tx-file FILEPATH       Input filepath of the JSON Tx.
+  --output-json            Format txid query output to JSON. Default format when
+                           writing to a file
+  --output-text            Format txid query output to TEXT. Default format when
+                           writing to stdout
```
smelc: This help text needs to be changed. Please don't review yet.

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -1723,6 +1723,22 @@ pOutputFormatJsonOrText kind =
...
+pTxIdOutputFormatJsonOrText =
+  asum
+    [ make OutputFormatJson "JSON" "json" False
+    , make OutputFormatText "TEXT" "text" True
+    ]
+ where
+  make format desc flag_ default_ =
+    Opt.flag' format $
+      mconcat
+        [ Opt.help $ "Format output to " <> desc <> (if default_ then " (the default)." else ".")
```
palas:
```suggestion
        [ Opt.help $ "Format output as " <> desc <> (if default_ then " (the default)." else ".")
```

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -1723,6 +1723,22 @@ pOutputFormatJsonOrText kind =
         , Opt.long ("output-" <> flag_)
         ]
 
+-- | @pTxIdOutputFormatJsonOrText kind@ is a parser to specify in which format
+-- to write @transaction txid@'s output on standard output.
+pTxIdOutputFormatJsonOrText :: Parser OutputFormatJsonOrText
+pTxIdOutputFormatJsonOrText =
+  asum
+    [ make OutputFormatJson "JSON" "json" False
+    , make OutputFormatText "TEXT" "text" True
```
palas: I think it would be best to have the default be a separate variable and compare to it, since we cannot ensure that there are no several `True` values in this list. Also the parser could already return the default instead of a `Maybe`, cause that would avoid having coupling with the implementation code.

palas (commented):
Looks good. I just have two suggestions.

Jimbo4350 (approved):
LGTM. Was this requested?

smelc:
> LGTM. Was this requested?

@Jimbo4350> no, I witnessed this was a possible improvement to consistency of our CLI interface when using the `id` command for another work.

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
smelc: > I think it would be best to have the default be a separate variable and compare to it, since we cannot ensure that there are no several True values in this list

@palas> done:

```diff
@@ -1728,14 +1728,15 @@ pOutputFormatJsonOrText kind =
 pTxIdOutputFormatJsonOrText :: Parser OutputFormatJsonOrText
 pTxIdOutputFormatJsonOrText =
   asum
-    [ make OutputFormatJson "JSON" "json" False
-    , make OutputFormatText "TEXT" "text" True
+    [ make OutputFormatJson "JSON" "json"
+    , make OutputFormatText "TEXT" "text"
     ]
  where
-  make format desc flag_ default_ =
+  default_ = OutputFormatText
+  make format desc flag_ =
     Opt.flag' format $
       mconcat
-        [ Opt.help $ "Format output as " <> desc <> (if default_ then " (the default)." else ".")
+        [ Opt.help $ "Format output as " <> desc <> (if format == default_ then " (the default)." else ".")
         , Opt.long ("output-" <> flag_)
         ]
 
diff --git a/cardano-cli/src/Cardano/CLI/Types/Common.hs b/cardano-cli/src/Cardano/CLI/Types/Common.hs
index 87253b0c7..d23ee97f2 100644
--- a/cardano-cli/src/Cardano/CLI/Types/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/Types/Common.hs
@@ -545,7 +545,7 @@ data TxMempoolQuery
 data OutputFormatJsonOrText
   = OutputFormatJson
   | OutputFormatText
-  deriving Show
+  deriving (Eq, Show)
```

> Also the parser could already return the default instead of a Maybe, cause that would avoid having coupling with the implementation code.

@palas> could we? Because we want the flags to be optional, to preserve the existing behavior.

palas: Yes, you can have a parser that returns the default on absence of a flag. I have done it somewhere, let me check

smelc: @palas> ah yes I could simply lift a value to `Parser` at the level of a `Parser (Maybe a)`. I have the same issue for https://github.com/IntersectMBO/cardano-cli/issues/1010 :slightly_smiling_face:

smelc: @palas> this works (for #1010):

```hs
pDRepIdOutputFormat :: Parser IdOutputFormat
pDRepIdOutputFormat =
  asum [make IdOutputFormatHex "hex", make IdOutputFormatBech32 "bech32"]
    <|> pure default_
 where
  default_ = IdOutputFormatBech32
  make format flag_ =
    Opt.flag' format $
      mconcat
        [ Opt.help $
            "Format drep id output as "
              <> flag_
              <> (if format == default_ then " (the default)." else ".")
        , Opt.long ("output-" <> flag_)
        ]
```

palas: That is it, I had forgotten the exact details

palas (approved):
Thanks 👍

Issue #883, opened 2024-09-05
[FEATURE] - extend drep id cmd so that it returns id in hex and bech32 for script based dreps. Use CIP129


CarlosLopezDeLara:
## What

Extend `cardano-cli conway drep id` so that it returns drep id in hex and bech32 also for script based dreps. 

~We should use Use [CIP105](https://cips.cardano.org/cip/CIP-0105)~ so that the prefix for script based dreps should be ~`drep_script`~

**EDIT:**
Community tools are moving towards CIP129. In particular, the prefix drep_script is deprecated in the latest changes to CIP105.  To ensure consistency with other tools we should adopt CIP129.  

CIP129 states that: 

> Binary format
In the header-byte, bits [7;4] indicate the type of gov key being used. The remaining four bits [3;0] are used to define the credential type. There are currently 3 types of credentials defined in the Cardano Conway era, this specification will allow us to define a maximum of 16 different types of keys in the future.
> 
> ```
>   1 byte     variable length   
>  <------> <-------------------> 
> ┌────────┬─────────────────────┐
> │ header │        key      │
> └────────┴─────────────────────┘
>     🔎                          
>     ╎          7 6 5 4 3 2 1 0  
>     ╎         ┌─┬─┬─┬─┬─┬─┬─┬─┐ 
>     ╰╌╌╌╌╌╌╌╌ |t│t│t│t│c│c│c│c│ 
>               └─┴─┴─┴─┴─┴─┴─┴─┘ 
> ```
> 

For the case of DReps we have: 

#### Key Type (Role) 
Key Type (`t t t t . . . .`)      | Key
---                               | ---
`0010....`                        | DRep

#### Credential Type

The second half of the header (bits [3;0]) refers to the credential type which can have the values and types as summarized in the table below,

Credential Type (`. . . . c c c c`)   | Semantic
---                                   | ---
`....0010`                            | Key Hash 
`....0011`                            | Script Hash 

So, for the DRep verification key

```
{
    "type": "DRepVerificationKey_ed25519",
    "description": "Delegate Representative Verification Key",
    "cborHex": "58208a10253789af22a6493dbe6a3a415732152989a9bd00b010e18f07a7cabcc74a"
}
```
we get the key-hash with:
```
cardano-cli conway governance drep id --drep-verification-key-file drep1.vkey --output-format hex
efda35608d806f37ac3948f1ddaff912e5f9130e9b52035c166bd8f2
```
[...]

PR #1009, opened 2025-01-08
drep id: support key hash

smelc:
# Changelog

```yaml
- description: |
    drep id: support key hash as input
  type:
```

# Context

While we're still unsure whether we should use https://cips.cardano.org/cip/CIP-0105 or https://cips.cardano.org/cip/CIP-129, we know we have to extend `drep id` to support key hashes and scripts. This PR makes a small forward and implements the former.

Part of fixing https://github.com/IntersectMBO/cardano-cli/issues/883

# How to trust this PR

Observe that we honor the behavior tagged as deprecated in CIP-105 (our existing already follows this behavior flagged "deprecated", so we are being consistent here):

image

* The existing golden tests for `drep id` still pass: we didn't change the default behavior

# Checklist

- [X] Self-reviewed the diff


smelc:
cc @CarlosLopezDeLara because this is the improvement we found out together during our pair programming session of yesterday

carbolymer (approved):

CarlosLopezDeLara (approved):

PR #1011, opened 2025-01-09
UTxO-HD targeting main

jasagredo:
# Changelog

```yaml
- description: |
    Implement changes needed for UTxO-HD Consensus feature.
    - Minor adjust on number of parameters of `AnyNewEpochState`.
  type:
```

# Context

See the [cardano-api PR](https://github.com/IntersectMBO/cardano-api/pull/719)

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,7 +13,7 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2024-12-24T12:56:48Z
+  , hackage.haskell.org 2025-01-14T03:16:27Z
   , cardano-haskell-packages 2025-01-15T09:59:24Z
 
 packages:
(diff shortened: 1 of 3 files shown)


cabal.project
```diff
@@ -32,11 +32,14 @@ package bitvec
   flags: -simd
 
 extra-packages: Cabal
+
 if os(windows)
   constraints: time ^>=1.14
   allow-newer: *:time
 
-constraints: any.text source
+-- Temporary workaround for problem with cabal-doctest
```
Jimbo4350: Points for the comment :+1:. Which repo is having issues with cabal-doctest?

Jimbo4350 (approved):

cabal.project
jasagredo: Ah that is from @neilmayhew's commits. Mine is just the last one. But I think it is consensus the one that had this problem. https://github.com/IntersectMBO/ouroboros-consensus/blob/main/cabal.project#L48-L49

neilmayhew: The problem occurred in api, cli and node as well as consensus. I think the problem arises because of the `extra-packages: Cabal` line. This allows a newer version of `Cabal` than would otherwise be chosen by the solver, which breaks `cabal-doctest`. However, there's been a release of `cabal-doctest` since I made this change, so it may be possible to drop it now.

neilmayhew: I tested this, and the constraint can be taken out in consensus, api and cli. However, it can't be taken out in node because it depends (indirectly) on `entropy` which uses a custom `Setup.hs` that isn't compatible with `Cabal-3.14` when built with `ghc8107`.

neilmayhew: Actually, that isn't correct. It's a haskell.nix problem. The cabal file for `entropy` has `setup-depends: Cabal >= 1.10 && < 3.13` and when building with cabal outside of haskell.nix it uses `Cabal-3.12.1.0` as expected, with both ghc8 and ghc9. However, haskell.nix uses `Cabal-3.14.1.0` for both.

However, this affects only node (⟶ `systemd` ⟶ `uuid` ⟶ `entropy`) so the constraint can be removed in consensus, api and cli.

It needs to be retained in node, but I'll push a commit that changes the comment.

neilmayhew: I've pushed a commit to my branch in cli that removes the constraint. You can just rebase on that.

neilmayhew: I've discovered that the problem can be worked around more cleanly in node by updating the Hackage index, because `entropy` was updated to be compatible with `Cabal-3.14` on Jan 1.

carbolymer (approved):

PR #1047, opened 2025-02-03
 Remove use of ScriptWitnessFiles in proposal scripts

Jimbo4350:
Depends on #1027 

# Changelog

```yaml
- description: |
    Remove use of ScriptWitnessFiles in proposal scripts

  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
@@ -21,6 +21,8 @@ import           Cardano.CLI.Environment (EnvCli (..), envCliAnyEon)
 import           Cardano.CLI.EraBased.Script.Certificate.Types (CliCertificateScriptRequirements)
 import qualified Cardano.CLI.EraBased.Script.Certificate.Types as Certifying
 import           Cardano.CLI.EraBased.Script.Mint.Types
+import           Cardano.CLI.EraBased.Script.Proposal.Types (CliProposalScriptRequirements)
+import qualified Cardano.CLI.EraBased.Script.Proposal.Types as Proposing
 import           Cardano.CLI.EraBased.Script.Spend.Types (CliSpendScriptRequirements)
 import qualified Cardano.CLI.EraBased.Script.Spend.Types as PlutusSpend
 import           Cardano.CLI.EraBased.Script.Vote.Types (CliVoteScriptRequirements)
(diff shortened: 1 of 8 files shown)


carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Script/Proposal/Read.hs
```diff
@@ -0,0 +1,120 @@
...
+        hoistIOEither $
+          readFileTextEnvelope AsProposal propFp
+  return (proposal, Nothing)
+readProposalScriptWitness w (propFp, Just certScriptReq) = do
+  let sbe = convert w
+  proposal <-
+    conwayEraOnwardsConstraints w $
+      modifyError (fmap TextEnvelopeError) $
+        hoistIOEither $
+          readFileTextEnvelope AsProposal propFp
```
palas, line 36:
```suggestion
  (proposal, _) <- readProposalScriptWitness w (propFp, Nothing)
```

cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs
```diff
@@ -1345,22 +1347,50 @@ pProposalFiles sbe balExUnits =
...
   pProposingScriptOrReferenceScriptWitness bExUnits =
-    pScriptWitnessFiles
-      WitCtxStake
+    pProposalScriptWitness
       bExUnits
       "proposal"
       Nothing
       "a proposal"
-      <|> pPlutusStakeReferenceScriptWitnessFilesVotingProposing "proposal-" balExUnits
+      <|> pProposalReferencePlutusScriptWitness "proposal-" balExUnits
```
palas: I would add the `-` in the inner function for consistency with `pProposalScriptWitness`. It also makes more sense to have the connectors in the function that makes the connection because it is only one place.

palas (approved):
Looks good to me. Just two lints

cardano-cli/src/Cardano/CLI/EraBased/Script/Proposal/Read.hs
Jimbo4350: This is more terse but I find it strange as we're saying there is no certificate script witness but there is.

PR #1077, opened 2025-03-04
Make it build with ghc-9.12

erikd:
# Changelog

```yaml
- description: Make it build with ghc-9.12
  type:
```

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.6", "9.8", "9.10"]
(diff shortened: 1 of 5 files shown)


.github/workflows/haskell.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.6", "9.8", "9.10"]
```
Jimbo4350, line 23: I see we only support `ghc: ["9.6", "9.10"]` in node. Can we drop `9.8`?

Jimbo4350 (approved):

PR #1094, opened 2025-03-22
Inline `subParser`

newhoggy:
# Changelog

```yaml
- description: |
    Inline `subParser`, `subParser'` and `hiddenSubParser` for more fine-grained control.
  type:
```

# Context

We currently use `subParser`, `subParser'` and `hiddenSubParser` for convenience.

The wrappers may be convenient, but they do not allow us to use to use available `optparse-applicative` combinators to control the rendering of CLI help easily.

Not using them affords us that flexibility.

# How to trust this PR

This PR is purely refactoring and produces no change in behaviour.  This is evidenced by no changes to the CLI help golden files.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Compatible/StakeAddress/Option.hs
+++ b/cardano-cli/src/Cardano/CLI/Compatible/StakeAddress/Option.hs
@@ -36,14 +36,15 @@ pStakeAddressRegistrationCertificateCmd
   => ShelleyBasedEra era
   -> Parser (CompatibleStakeAddressCmds era)
 pStakeAddressRegistrationCertificateCmd sbe = do
-  subParser "registration-certificate" $
-    Opt.info
-      ( CompatibleStakeAddressRegistrationCertificateCmd sbe
-          <$> pStakeIdentifier Nothing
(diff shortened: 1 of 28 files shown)


cardano-cli/src/Cardano/CLI/Compatible/StakeAddress/Option.hs
```diff
@@ -36,14 +36,15 @@ pStakeAddressRegistrationCertificateCmd
...
-  subParser "registration-certificate" $
-    Opt.info
-      ( CompatibleStakeAddressRegistrationCertificateCmd sbe
-          <$> pStakeIdentifier Nothing
-          <*> pFeatured (toCardanoEra sbe) pKeyRegistDeposit
-          <*> pOutputFile
-      )
-      desc
+  Opt.hsubparser $
+    commandWithMetavar "registration-certificate" $
```
palas, line 40: We could have a `subCommandWithMetavar` since we are at it

palas (approved):
Looks good, just an idea.

Also, I am a bit surprised that the golden files don't change at all given that `Opt.hsubparser` supposedly adds `--help`, but I guess we had it already?

carbolymer (approved):

PR #1101, opened 2025-03-25
New `CARDANO_CLI_VISIBILITY_LEVEL` environment variable

newhoggy:
# Changelog

```yaml
- description: |
    Add a new `CARDANO_CLI_VISIBILITY_LEVEL` environment variable which can be defined as `internal`, `hidden` or `visible`. 
  type:
```

# Context

It is useful to be able to see what the hidden and internal commands are even though they are hidden by default.

The feature is also used to generate golden files that include CLI help for hidden and internal commands so that hidden and internal APIs are also tracked.

# How to trust this PR

No changes to the golden files apart from the rename of `help.cli` to `help-visible.cli` and the addition of `help-hidden.cli` and `help-internal.cli` files.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/Run.hs
@@ -53,7 +53,9 @@ import Data.Text.IO qualified as Text
 import Data.Version (showVersion)
 import Options.Applicative.Help.Core
 import Options.Applicative.Types
-  ( OptReader (..)
+  ( OptProperties (..)
+  , OptReader (..)
+  , OptVisibility (..)
   , Option (..)
   , Parser (..)
   , ParserInfo (..)
@@ -113,8 +115,8 @@ runClientCommand = \case
     firstExceptT PingClientError $ runPingCmd cmds
   CliDebugCmds cmds ->
     firstExceptT DebugCmdError $ runDebugCmds cmds
-  Help pprefs allParserInfo ->
-    runHelp pprefs allParserInfo
+  Help visibility pprefs allParserInfo ->
+    runHelp visibility pprefs allParserInfo
   DisplayVersion ->
     runDisplayVersion
 
@@ -161,8 +163,8 @@ runDisplayVersion = do
  where
   renderVersion = Text.pack . showVersion
 
-helpAll :: ParserPrefs -> String -> [String] -> ParserInfo a -> IO ()
-helpAll pprefs progn rnames parserInfo = do
+helpAll :: OptVisibility -> ParserPrefs -> String -> [String] -> ParserInfo a -> IO ()
+helpAll visibility pprefs progn rnames parserInfo = do
   IO.putStrLn $ customRenderHelp 80 (usage_help parserInfo)
   IO.putStrLn ""
   go (infoParser parserInfo)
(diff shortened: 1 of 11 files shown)


Jimbo4350 (commented):
Where will this be used?

cardano-cli/test/cardano-cli-golden/Test/Golden/Help.hs
```diff
@@ -63,7 +63,47 @@ hprop_golden_HelpAll =
...
+    -- These tests are not run on Windows because the cardano-cli usage
+    -- output is slightly different on Windows.  For example it uses
+    -- "cardano-cli.exe" instead of "cardano-cli".
+    unless isWin32 $ do
+      helpFp <- H.note "test/cardano-cli-golden/files/golden/help-internal.cli"
+
+      help <-
+        filterAnsi
+          <$> execCardanoCLIWithEnvVars
+            [ ("CARDANO_CLI_VISIBILITY_LEVEL", "internal")
```
newhoggy, line 105: It's used here.

We ensure that `hidden` and `internal` cli commands are still tracked by golden files so we don't break them accidentally.

cardano-cli/src/Cardano/CLI/Run.hs
```diff
@@ -161,20 +163,22 @@ runDisplayVersion = do
...
   go :: Parser a -> IO ()
   go p = case p of
     NilP _ -> return ()
-    OptP optP -> case optMain optP of
-      CmdReader _ cs -> do
-        forM_ cs $ \(c, subParserInfo) ->
-          helpAll pprefs progn (c : rnames) subParserInfo
-      _ -> return ()
+    OptP optP ->
+      let visible = propVisibility (optProps optP) >= visibility
```
palas, line 176: From the point of view of a user facing parameter, having 3 values is what makes most sense, indeed.

But from the point of view of the golden files, an issue I see is that `help-internal.cli` includes `help-hidden.cli` which includes `help.cli`, which is a lot of repeated stuff.

And idea is we could have something like unix permissions, so that, for example:
- 1 = visible (x)
- 2 = hidden (w)
- 3 = visible + hidden (wx)
- 4 = internal (r)
...
- 7 = everything  (rwx)

A bit less intuitive though. But could still also accept "visible" and interpret as 1, etc. Could even have "only_hidden" as 2, "not_internal" as 3, or "all" as 7.

cardano-cli/test/cardano-cli-golden/Test/Golden/Help.hs
```diff
@@ -97,7 +137,9 @@ hprop_golden_HelpCmds =
     unless isWin32 $ do
       help <-
         filterAnsi
-          <$> execCardanoCLI
+          <$> execCardanoCLIWithEnvVars
+            [ ("CARDANO_CLI_VISIBILITY_LEVEL", "internal")
+            ]
```
palas, line 142: Could also use the opportunity to tag the files as `internal` or `hidden` (maybe putting them in different folders). Better rebase first though, cause I just modified that bit of code

palas (commented):
I think visibilising what is internal/hidden and what is not, in the golden files, is very positive. But I think it would be even better to have them categorised instead of repeated.

Jimbo4350 (changes requested):
There are 5 uses of `hidden` in `cardano-cli` (but many more calls to `internal`) . The vast majority are for deprecated commands (i.e they won't change) and actually should be removed. Let's remove all of the explicitly deprecated options first. Then get feedback for the options not explicitly marked deprecated. I suspect all of the `internal`/`hidden` options are not in use anymore and we can likely get rid of them.

PR #1118, opened 2025-04-01
Rename `AllOutputFormats` to `LedgerOutputFormat`

newhoggy:
# Changelog

```yaml
- description: |
    Rename `AllOutputFormats` to `LedgerOutputFormat`
  type:
  # maintenance    # not directly related to the code
```

# Context

The name `AllOutputFormats` is misleading because it doesn't actually represent all possible output formats.

Inspecting the code, it is limited to those outputs that are representations of ledger state (in that they have CBOR representation) so the new name is chosen accordingly.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Type/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/Type/Common.hs
@@ -7,7 +7,7 @@
 
 module Cardano.CLI.Type.Common
   ( AllOrOnly (..)
-  , AllOutputFormats (..)
+  , LedgerOutputFormat (..)
   , AddressKeyType (..)
   , AnchorScheme (..)
   , AnyPlutusScriptVersion (..)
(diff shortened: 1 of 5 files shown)


carbolymer (commented):
I'm not sure the name is better. What's so "ledger" about the output? It's only used in dumping UTxOs, so why not `UtxoOutputFormat`?

This PR feels like it's a part of a larger change and the context is missing here.

newhoggy:
I was assume it was the case that all the CBOR stuff is CBOR so it goes onto the ledger.

As for `UtxoOutputFormat`, I think it is too specific.  It seems a lot of define a type for each possible output, and the original name `AllOutputFormats` seems wrong.  There are possibly other commands that output things other than Utxos that could share the same type.

The context for the PR is that I'm working on https://github.com/IntersectMBO/cardano-cli/issues/566 and wanted to do some refactoring as I familiarise myself with the code.  Part of that task would be to decide what type I need to use to represent the possible output formats for each command and having more intuitive names for the format types would help me choose the right types.

cardano-cli/src/Cardano/CLI/Type/Common.hs
```diff
@@ -489,10 +489,12 @@ data OutputFormatJsonOrText
   | OutputFormatText
   deriving (Eq, Show)
 
-data AllOutputFormats
-  = FormatJson
-  | FormatText
-  | FormatCBOR
+-- | The output format for ledger outputs.  Ledger outputs are those outputs that have a ledger representation
+-- (ie. CBOR).
+data LedgerOutputFormat
```
Jimbo4350, line 494: We need to redesign this and apply it to all commands. Do you consider the `TextEnvelope` format to be a ledger format?

PR #1132, opened 2025-04-08
Skip HLS on trivial changes

palas:
# Changelog

```yaml
- description: |
    Skip HLS checks on trivial changes
  type:
```

# Context

Doing releases doesn't affect whether HLS works or not. For that reason, we want to skip the check when it doesn't make sense. This PR introduces an exception for when only markdown is changed and for when only the version in a cabal file is changed.

# How to trust this PR

That is tricky. I've been testing the different possibilities, but it isn't easy to test.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -21,7 +21,59 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Check if changes were trivial
+        id: check_trivial_changes
+        run: |
+          git fetch origin ${{ github.base_ref }} --unshallow
+          base_ref=origin/${{ github.base_ref }}
+          head_ref=HEAD
+          changed_files=$(git diff-tree --name-status -r "$base_ref".."$head_ref" -- | cut -f2 -d$'\t')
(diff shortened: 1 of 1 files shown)


.github/workflows/hls.yml
```diff
@@ -21,24 +21,80 @@ jobs:
...
+            else
+              # If other types of files were changed, do not skip the checks
+              echo "$file: was changed and is not a markdown nor a cabal file (non-trivial change)"
+              exception=false
+              break
+            fi
+          done
+
+          if $exception; then
+            echo "RUN_TESTS=0" >> "$GITHUB_OUTPUT"
```
Jimbo4350: No we still want to run tests.

Jimbo4350 (changes requested):

Jimbo4350 (approved):

.github/workflows/hls.yml
palas: Renamed the variable as agreed. Now it is called `CHECK_HLS_WORKS`, which is more intuitive

carbolymer (approved):

PR #1184, opened 2025-05-14
Simplify by not using `QueryStakeAddressInfoCmdArgs` unnecessarily

newhoggy:
# Changelog

```yaml
- description: |
    Simplify code by not using `QueryStakeAddressInfoCmdArgs` unnecessarily.
    Also fixes `query spo-stake-distribution` command so that the output format flags are respected.
  type:
```

# Context

Don't build a value of a data type if it doesn't make sense to use all the fields.  Just pass the values directly to the function instead.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
@@ -962,12 +962,13 @@ runQueryStakeAddressInfoCmd
   => Cmd.QueryStakeAddressInfoCmdArgs
   -> ExceptT QueryCmdError IO ()
 runQueryStakeAddressInfoCmd
-  cmd@Cmd.QueryStakeAddressInfoCmdArgs
+  Cmd.QueryStakeAddressInfoCmdArgs
     { Cmd.commons =
-      Cmd.QueryCommons
+      commons@Cmd.QueryCommons
         { Cmd.nodeConnInfo
         , Cmd.target
         }
+    , Cmd.addr
     , Cmd.outputFormat
     , Cmd.mOutFile
     } = do
@@ -978,7 +979,7 @@ runQueryStakeAddressInfoCmd
         & onLeft (left . QueryCmdUnsupportedNtcVersion)
     sbe <- requireShelleyBasedEra era & onNothing (left QueryCmdByronEra)
 
-    said <- callQueryStakeAddressInfoCmd cmd
+    said <- callQueryStakeAddressInfoCmd commons addr
 
     writeStakeAddressInfo sbe said outputFormat mOutFile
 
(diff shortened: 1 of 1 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1835,19 +1832,11 @@ runQuerySPOStakeDistribution
               | (keyHash, addr) <- Map.toList $ L.psStakePoolParams poolState
               ]
 
-        mkQueryStakeAddressInfoCmdArgs addr =
-          Cmd.QueryStakeAddressInfoCmdArgs
-            { Cmd.commons = commons
-            , addr
-            , outputFormat
-            , mOutFile -- unused anyway. TODO tighten this by removing the field.
```
newhoggy, line 1843: Both `outputFormat` and `mOutFile` are unused by `callQueryStakeAddressInfoCmd`, so don't bother using `QueryStakeAddressInfoCmdArgs` and packing unnecessary fields.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1835,19 +1832,11 @@ runQuerySPOStakeDistribution
...
-            }
-
     spoToDelegatee <-
       Map.fromList . concat
         <$> traverse
           ( \stakeAddr -> do
-              info <- callQueryStakeAddressInfoCmd $ mkQueryStakeAddressInfoCmdArgs stakeAddr
+              info <- callQueryStakeAddressInfoCmd commons stakeAddr
```
newhoggy: Instead call `callQueryStakeAddressInfoCmd` with the values it needs directly.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1801,7 +1799,6 @@ runQuerySPOStakeDistribution
         , Cmd.target
         }
     , Cmd.spoHashSources = spoHashSources'
-    , Cmd.outputFormat
```
newhoggy: The unnecessary use of `QueryStakeAddressInfoCmdArgs` hides problems.  For example here `outputFormat` is actually unused, when it really should be.  The use of `QueryStakeAddressInfoCmdArgs` hides this fact.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1856,15 +1846,26 @@ runQuerySPOStakeDistribution
...
           [ ( spo
             , coin
             , Map.lookup spo spoToDelegatee
             )
           | (spo, coin) <- Map.assocs spoStakeDistribution
           ]
 
-    writeOutput mOutFile toWrite
+        output =
+          outputFormat
```
newhoggy, line 1858: We now respect the output format of the `query spo-stake-distribution` command.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1856,15 +1846,26 @@ runQuerySPOStakeDistribution
...
+            & ( id
```
palas, line 1859: Is this `id` to get alignment from the formatter? I don't think it has any effect on the code

newhoggy: Yes, it is purely for alignment purposes.

Jimbo4350 (approved):

PR #1227, opened 2025-06-25
Delete `watchdogProp` function

newhoggy:
# Changelog

```yaml
- description: |
    Delete `watchdogProp` function
  type:
```

# Context

`watchdogProp` can be deleted because we are now using a new version of `hedgehog-extras` that fixes a hanging bug in `moduleWorkspace`.  After fixing this bug, tests do not hang anymore so we do not need `watchdogProp` to interrupt the hanging test.

The PR that fixed `moduleWorkspace` is here: https://github.com/input-output-hk/hedgehog-extras/pull/91

The reason the old version of `moduleWorkspace` hangs is not fully known, however it uses a combination of `ResourceT` and `recovering retryPolicy`, which causes the hang.

Aside from the fact that this combination hangs, it is problematic to use `ResourceT` when the intention is to cleanup resources in a timely manner.  The scope of `ResourceT` is `runResourceT` which is outside the call to `moduleWorkspace`.  This means that when `moduleWorkspace` returns, there is no guarantee that the module directory is cleaned up as it is supposed to.

In order to ensure timely cleanup of resources, the `bracket` family of functions should be used instead.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,7 +13,7 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2025-06-22T20:18:27Z
+  , hackage.haskell.org 2025-06-27T07:48:39Z
   , cardano-haskell-packages 2025-06-22T22:27:17Z
 
 packages:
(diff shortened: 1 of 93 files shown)


Jimbo4350 (commented):
Can you add an explanation in the description as to why we can delete `watchdogProp`?

Jimbo4350 (approved):

carbolymer (changes requested):
https://github.com/IntersectMBO/cardano-cli/pull/1225#pullrequestreview-2962876334

PR #1235, opened 2025-06-29
Use everything under `Hedgehog.Extras.Test` from `hedgehog-extras` via module `Test.Cardano.CLI.Hedgehog`

newhoggy:
# Changelog

```yaml
- description: |
    Use everything under `Hedgehog.Extras.Test` from `hedgehog-extras` via module `Test.Cardano.CLI.Hedgehog`
  type:
```

# Context

Having all the tests `Test.Cardano.CLI.Hedgehog` instead of `hedgehog-extras` directly means some functions can be replaced with local versions which can aid in the testing of different versions of the local version before they are upstreamed to `hedgehog-extras`.

This means we can iterate quickly on debugging the hanging problem by substituting our own version of `moduleWorkspace` `runWatchdog` and similar functions.

Doing it this way saves us from having to release `hedgehog-extras` without being sure the problem is solved just so that it can be tested in CI.

Once we are sure the hanging problem is resolved we can upstream the fixes and if we wish rollback the introduction of this module.

# How to trust this PR

Highlight important bits of the PR that will make the review faster. If there are commands the reviewer can run to observe the new behaviour, describe them.

# Checklist

- [ ] Self-reviewed the diff


carbolymer (commented):
What's the benefit here?

PR #1236, opened 2025-06-29
Local error golden test functions

newhoggy:
# Changelog

```yaml
- description: |
    Import from `cardano-api` into local functions:
    * `testAllErrorMessages`
    * `testAllErrorMessages_`
    *  `testErrorMessage`
    Call `watchdogProp` when running golden error tests
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -31,6 +31,14 @@ package bitvec
   -- Workaround for windows cross-compilation
   flags: -simd
 
+package asn1-encoding
+  -- This is needed because asn1-encoding fails to build on Windows in CI due to it generating warnings.
+  ghc-options: -Wno-unused-imports
+               -Wno-noncanonical-monad-instances
+               -Wno-x-partial
+               -Wno-name-shadowing
+               -Wno-incomplete-uni-patterns
+
 extra-packages: Cabal
 if os(windows)
   constraints: time ^>=1.14
--- a/cardano-cli/cardano-cli.cabal
+++ b/cardano-cli/cardano-cli.cabal
@@ -318,6 +318,7 @@ library cardano-cli-test-lib
   hs-source-dirs: test/cardano-cli-test-lib
   exposed-modules:
     Test.Cardano.CLI.Aeson
+    Test.Cardano.CLI.Golden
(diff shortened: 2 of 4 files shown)


cabal.project
```diff
@@ -31,6 +31,13 @@ package bitvec
   -- Workaround for windows cross-compilation
   flags: -simd
 
+package asn1-encoding
+  ghc-options: -Wno-unused-imports
+               -Wno-noncanonical-monad-instances
+               -Wno-x-partial
+               -Wno-name-shadowing
+               -Wno-incomplete-uni-patterns
+
```
erikd, line 41: Why is this needed? Would be nice to have a comment explaining why.

newhoggy: One of the Windows builds fails otherwise.  The `asn1-encoding` package fails to build because it generates warnings on Windows and our build errors out on warnings.

newhoggy: Comment added.

erikd (approved):

Jimbo4350 (dismissed):
Why are you moving the functions here?

carbolymer (commented):
I have the same question as @Jimbo4350

carbolymer (dismissed):
I have the same question as @Jimbo4350 . Why vendor in functions here, rather than fixing them upstream?

I'm not convinced this PR is a net benefit because I haven't seen hangups in golden tests in `cardano-api` and we are using those functions there as well.

@newhoggy Have you seen hangups in error messages' goldens? Have you seen them in cardano-api as well?

newhoggy:
The reason I want this is there are unexplained hangs and if watchdogProp can catch it I can log which test it is and there was one hang where this test was the last test that ran.

If `watchdogProp` doesn't wrap a test and I run a test and it hangs, I can't tell if it's one of the unwrapped tests.

The PR is not meant to be permanent.  When the root cause of the hangs are determined, this can be rolled back and any changes upstreamed.

newhoggy:
We know the hangs are due to a GHC bug, so this can be closed.

PR #1250, opened 2025-07-15
Move protocol parameters update code

Jimbo4350:
# Changelog

```yaml
- description: |
    Move protocol parameters update code
  type:
  - refactoring   
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Governance/Actions/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Governance/Actions/Run.hs
@@ -23,6 +23,7 @@ import Cardano.Api.Ledger (StrictMaybe (..))
 import Cardano.Api.Ledger qualified as L
 
 import Cardano.CLI.Compatible.Exception
+import Cardano.CLI.Compatible.Governance.Types
 import Cardano.CLI.Compatible.Json.Friendly
 import Cardano.CLI.EraBased.Governance.Actions.Command
 import Cardano.CLI.EraBased.Governance.Actions.Command qualified as Cmd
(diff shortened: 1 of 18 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Governance/Actions/Run.hs
```diff
@@ -302,90 +304,51 @@ runGovernanceActionCreateProtocolParametersUpdateCmd
...
+        UpdatePParams
+          prevGovActId
+          updateProtocolParams
+          (toShelleyScriptHash <$> L.maybeToStrictMaybe mConstitutionalScriptHash)
+
+  let proposalProcedure = createProposalProcedure sbe network deposit depositStakeCredential govAct proposalAnchor
+
+  fromEitherIOCli @(FileError ()) $
+    obtainCommonConstraints era $
+      writeFileTextEnvelope oFp (Just "Update protocol parameters proposal") proposalProcedure
```
palas, line 346: It seems this is really close to `conwayProtocolParametersUpdate` now. Would it be possible to use `conwayProtocolParametersUpdate` here instead? Maybe moving it to this module. Or is there a reason why that is a bad idea that I am missing?

palas (approved):
Looks good, just a suggestion to try to avoid replication 👍

cardano-cli/src/Cardano/CLI/EraBased/Governance/Actions/Run.hs
Jimbo4350: It's the same but the plan is to further abstract this to make it clear it's parameterized on `Era era` and reuse it in the compatiblility functions.

PR #1324, opened 2026-01-20
Improve error message for malformed timestamps

### LINKED ISSUE #1166: query slot-number <malformated-timestamp> exits with CallStack error
using a malformated timestamp like
`cardano-cli conway query slot-number 123456`
exits with error:
```
cardano-cli: parseTimeOrError: no parse of "123456"
CallStack (from HasCallStack):
  error, called at libraries/time/lib/Data/Time/Format/Parse.hs:124:15 in time-1.12.2:Data.Time.Format.Parse
```

```console
cardano-cli version
cardano-cli 10.8.0.0 - linux-x86_64 - ghc-9.6
git rev 85b6a871d57bd4789283283d17451c10f74e84d2
```

palas:
# Changelog

```yaml
- description: |
    Improved error message for malformed timestamps
  type:
  - bugfix
```

# Context

See https://github.com/IntersectMBO/cardano-cli/issues/1166.

# How to trust this PR

It is reassuring that the golden files for the parameters didn't change, so the flags didn't change. The thing to check is whether the error is correct, and whether it parses normally when provided a right output. You can use the example in the original issue:

```bash
cardano-cli conway query slot-number 123456
```

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Common/Option.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Common/Option.hs
@@ -51,7 +51,7 @@ import Data.Maybe
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Time.Clock (UTCTime)
-import Data.Time.Format (defaultTimeLocale, parseTimeOrError)
+import Data.Time.Format (defaultTimeLocale, parseTimeM)
 import Data.Word
 import GHC.Exts (IsList (..))
 import GHC.Natural (Natural)
@@ -59,6 +59,7 @@ import Lens.Micro
 import Network.Socket (PortNumber)
 import Options.Applicative hiding (help, str)
 import Options.Applicative qualified as Opt
+import Options.Applicative.Types (readerAsk)
 import Text.Read (readEither, readMaybe)
 import Text.Read qualified as Read
 import Vary (Vary, (:|))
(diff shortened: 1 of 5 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Common/Option.hs
```diff
@@ -1413,9 +1414,19 @@ pTxMetadataJsonSchema =
...
-  parseTimeOrError False defaultTimeLocale "%Y-%m-%dT%H:%M:%SZ"
+convertTime :: String -> Maybe UTCTime
+convertTime = parseTimeM False defaultTimeLocale "%Y-%m-%dT%H:%M:%SZ"
+
+timeReader :: ReadM UTCTime
+timeReader = do
+  arg <- readerAsk
+  maybe
+    ( readerError $
+        "Malformed timestamp `" ++ arg ++ "', use UTC timestamp in YYYY-MM-DDThh:mm:ssZ format."
```
carbolymer, line 1425: You've used different beginning quote and end quote for the argument. This may look odd.

carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Common/Option.hs
palas: Well spotted, and you do well in pointing it out, but it is the default for the `optparse-applicative` library, I copied it from here: https://hackage-content.haskell.org/package/optparse-applicative-0.19.0.0/docs/src/Options.Applicative.Builder.html#maybeReader
So it seems the most appropriate

PR #1341, opened 2026-02-25
Fix supplemental datum propagation in transaction build

Jimbo4350:
# Changelog

```yaml
- description: |
    Fix supplemental datum propagation in transaction outputs
    Previously, supplemental datums (datum hashes referenced in transaction outputs) were not being propagated into the transaction body. This meant that when a transaction output included a datum hash, the corresponding datum was not included in the transaction's supplemental data map, causing validation failures for consumers expecting to resolve those datum hashes.
  type:
    - bugfix     
```


carbolymer:
Your PR is not using description template

carbolymer (approved):

PR #1345, opened 2026-03-06
Fix ping command outputting garbage JSON on failure

### LINKED ISSUE #1210: ping command outputs garbage upon failure
When running the following

```
cardano-cli ping  -c1 -q -j -h cardano.hydra.bzh -p 3002 2>/dev/null
```

and the `ping` command is unable to connect to the host:port, I get this output:

```
] }
```

which does not really make any sense. This annoying because it can break script pipelines trying to extract information from the `ping` command output.

Ideally, I would like the output to be consistent whether it's a success or a failure.

Jimbo4350:
# Changelog

```yaml
- description: |
    Fix ping command outputting invalid JSON ("] }") when connection fails
  type:
    - bugfix
```

# Context

When running `cardano-cli ping -c1 -q -j -h <host> -p <port>` and the connection fails, the output was:

```
] }
```

This is because the upstream `CNP.logger` prints `] }` on `LogEnd` even when no opening brace was emitted (no `LogMsg` was ever received). This breaks script pipelines expecting valid JSON.

This PR replaces `CNP.logger` with a local version that:
- Outputs a complete empty JSON object (`{ "pongs": [] }` or `{ "tip": [] }`) when no messages were logged
- Behaves identically to the upstream logger otherwise

Closes #1210

# Checklist

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraIndependent/Ping/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/EraIndependent/Ping/Run.hs
@@ -17,9 +17,10 @@ import Cardano.Network.Ping qualified as CNP
 import Control.Concurrent.Class.MonadSTM.Strict (StrictTMVar)
 import Control.Concurrent.Class.MonadSTM.Strict qualified as STM
 import Control.Exception (SomeException)
-import Control.Monad (forM, unless)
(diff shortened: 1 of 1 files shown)


Jimbo4350:
Closing in favour of fixing this upstream in cardano-ping's `logger` function.

PR #1375, opened 2026-05-12
Remove old-API TxBody uses from cardano-cli (companion to cardano-api #1200)

Jimbo4350:
## Summary

Companion to cardano-api [#1200](https://github.com/IntersectMBO/cardano-api/pull/1200), which deprecates the old transaction-body API surface. This PR strips the remaining old-API uses from cardano-cli so `-Werror` stays clean once cardano-cli bumps to a cardano-api containing #1200.

PR [#1374](https://github.com/IntersectMBO/cardano-cli/pull/1374) already migrated the friendly-tx renderer; this PR handles the rest: `Cardano.CLI.Read`, `Cardano.CLI.EraBased.Transaction.Run`, `Cardano.CLI.EraIndependent.Debug.TransactionView.Run`, and the `TxBodyFile` phantom alias in `Cardano.CLI.Type.Common`.

Deprecated symbols replaced:

| Old (deprecated by #1200) | Replacement |
|---|---|
| `TxBody era` | `Exp.UnsignedTx (ShelleyLedgerEra era)` |
| `getTxBody` | pattern-match `ShelleyTx _ ledgerTx` |
| `getTxBodyAndWitnesses` | `getTxWitnesses` + ledger body via lens |
| `makeShelleyKeyWitness` | `makeShelleyKeyWitness'` (ledger body) |
| `makeSignedTransaction` | `addWitnesses` |
| `getTxId . getTxBody` | `getTxIdShelley sbe (ledgerTx ^. L.bodyTxL)` |
| `evaluateTransactionExecutionUnits` (on old `TxBody`) | `evaluateTransactionExecutionUnitsShelley` (ledger `Tx`) |
| `evaluateTransactionFee` (on old `TxBody`) | `Exp.evaluateTransactionFee` (on `Exp.UnsignedTx`) |

Pre-Conway era support is preserved everywhere it currently works. The migration uses ledger-level signing (`makeShelleyKeyWitness'` / `addWitnesses`) rather than `Exp.makeKeyWitness` / `Exp.signTx`, which are Conway+ only. The TextEnvelope I/O boundary still uses `Tx era` / `ShelleyTx _` (both non-deprecated); `.tx`/`.txbody` on-disk formats are unchanged.

The compatible command path (`Cardano.CLI.Compatible.Transaction.Run`) already used non-deprecated symbols and is unchanged.

## Commits

Reviewable in order:

[...]

PR #1376, opened 2026-05-14
Migrate to experimental TxOut in createCompatibleTx

Jimbo4350:
# Changelog

```yaml
- description: |
    Migrate `createCompatibleTx`, `toTxOutInEra`, and `toTxOutInShelleyBasedEra`
    to the experimental `TxOut` plus supplemental-datum map introduced by
    cardano-api PR #1209. Drop `TxCmdDatumDecodingError` (the underlying
    `Exp.DatumDecodingError` is deleted upstream).
  type:
   - maintenance
```

# Context

Depends on **cardano-api [#1209](https://github.com/IntersectMBO/cardano-api/pull/1209)** — _Remove legacy TxOut from Compatible and Experimental APIs_.

That PR makes two breaking changes in the cardano-api experimental namespace:

1. `createCompatibleTx` now takes `[Exp.TxOut (ShelleyLedgerEra era)]` plus a new `Map L.DataHash (L.Data ...)` of supplemental datum bodies. The legacy `TxOut CtxTx era` bundled supplemental datums inside outputs; `Exp.TxOut` only carries the datum hash, so callers thread the full datum bodies in explicitly.
2. The bridge helpers `fromLegacyTxOut`, `legacyDatumToDatum`, `supplementalDatumFromLegacy`, `toLedgerDatum`, and the `DatumDecodingError` type are removed from `Cardano.Api.Experimental.Tx`.

This PR adapts cardano-cli's three affected call sites and removes the now-unreachable error constructor.

## How the migration works

A new helper `convertLegacyTxOut :: ShelleyBasedEra era -> TxOut CtxTx era -> (Exp.TxOut (ShelleyLedgerEra era), Map DataHash (L.Data ...))` lives in `Cardano.CLI.Compatible.Transaction.TxOut`. It builds on the existing `toShelleyTxOutAny` (which already strips supplemental datums to their hash in the ledger TxOut) and pulls the supplemental datum bodies into the map. Total conversion — no error case, so the previous `Either DatumDecodingError ...` plumbing is gone.

Call sites:

- `Cardano.CLI.Compatible.Transaction.Run`: converts outs at the boundary and folds supplemental datums via `Map.unions` before calling `createCompatibleTx`.
- `Cardano.CLI.EraBased.Transaction.Run`: `toTxOutInEra` and `toTxOutInShelleyBasedEra` use `convertLegacyTxOut` instead of the deleted `Exp.fromLegacyTxOut`. `obtainCommonConstraints` bridges `ShelleyLedgerEra ~ LedgerEra` for the experimental-era return type.

## Temporary `cabal.project` additions

To let CI validate against the unpublished API:

- A `source-repository-package` pointing at the cardano-api PR #1209 branch.
- Per-package `-Wwarn=deprecations -Wwarn=unused-imports` for cardano-cli.

The PR branch is several commits ahead of `cardano-api-11.1.0.0` and surfaces unrelated `TxBody`/`TxBodyContent` deprecations (cardano-api [#1200](https://github.com/IntersectMBO/cardano-api/pull/1200)) plus a redundant `Cardano.Ledger.Core` import in `Cardano.CLI.Read` that became visible after upstream re-exports widened. Both are separate cardano-cli cleanups; this PR keeps them out of scope by downgrading the warnings. **All four lines should be removed in a follow-up once #1209 is merged and the next cardano-api is published to CHaP.**

# How to trust this PR

- The migration is mechanical: one new helper, three replaced call sites, one removed error constructor.
- `convertLegacyTxOut` reuses `toShelleyTxOutAny` rather than re-implementing the datum-to-ledger mapping, so semantics on the `Exp.TxOut` side are identical to what cardano-api itself does for non-experimental tx body construction.
- Supplemental datums are still threaded into the witness set: the helper returns them in the map, and the new `createCompatibleTx` argument is passed through to `convScriptData'` inside cardano-api.
- Build is clean (lib + exe + both test suites) against the SRP'd cardano-api branch with `-Werror` still in effect for everything except the two pre-existing classes of warnings above.

# Checklist

- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details — _no tests added; behavior on the cardano-cli surface is unchanged_
- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -26,6 +26,29 @@ packages:
 program-options
   ghc-options: -Werror
 
+-- TEMPORARY: build against cardano-api PR #1209
+-- (jordan/remove-legacy-txout-from-compat-and-experimental).
+-- The PR removes legacy TxOut from the Compatible and Experimental APIs,
+-- which this branch migrates cardano-cli to. Remove this stanza and the
+-- per-package warning downgrades below once #1209 is merged and the next
+-- cardano-api release is published to CHaP.
+source-repository-package
+    type: git
+    location: https://github.com/IntersectMBO/cardano-api
+    tag: d848cec1ba4769b9daf4ac5ac9ed61a3f76c8042
+    subdir:
+      cardano-api
+      cardano-api-gen
+
+-- TEMPORARY: the PR branch is several commits ahead of cardano-api-11.1.0.0
+-- and surfaces unrelated TxBody/TxBodyContent deprecations (cardano-api PR
+-- #1200) plus a redundant Cardano.Ledger.Core import in Cardano.CLI.Read
+-- that became visible after upstream re-exports widened. Both are separate
+-- cardano-cli cleanups; downgrade to warnings here so the migration build
+-- is not blocked by them. Remove once the underlying cleanups land.
+package cardano-cli
+  ghc-options: -Wwarn=deprecations -Wwarn=unused-imports
+
 package crypton
   -- Using RDRAND instead of /dev/urandom as an entropy source for key
   -- generation is dubious. Set the flag so we use /dev/urandom by default.
--- a/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
+++ b/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
@@ -33,6 +33,7 @@ where
 import Cardano.Api as Api
 import Cardano.Api.Experimental (obtainCommonConstraints)
 import Cardano.Api.Experimental qualified as Exp
+import Cardano.Api.Experimental.Tx qualified as Exp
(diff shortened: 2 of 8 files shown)


Jimbo4350:
Superseded by #1379, which folds in this PR's work (cardano-api #1209 companion — migration of `createCompatibleTx`, `toTxOutInEra`, and `toTxOutInShelleyBasedEra` to the experimental `TxOut` plus supplemental datum map) along with the cardano-api #1200 `TxBody` deprecation migration (was #1375) and the cardano-api #1210 Certificate API migration. All three now consume the published cardano-api 11.2.0.0 from CHaP rather than a source-repository-package pin. Closing in favour of #1379.

PR #1387, opened 2026-06-01
Minor updates

erikd:
# Context
Minior updates:
* Update index-states
* Remove allow-newers
* Update Nix

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,8 +13,8 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2026-04-16T00:15:23Z
-  , cardano-haskell-packages 2026-05-26T09:18:49Z
+  , hackage.haskell.org 2026-06-02T21:49:32Z
+  , cardano-haskell-packages 2026-06-02T21:14:32Z
 
 active-repositories:
   , :rest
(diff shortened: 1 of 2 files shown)


cabal.project
```diff
@@ -74,69 +74,33 @@ write-ghc-environment-files: always
 -- Do NOT add more source-repository-package stanzas here unless they are strictly
 -- temporary! Please read the section in CONTRIBUTING about updating dependencies.
 
+-- ouroboros-consensus currently needs this. Version 1.2 has an breaking API change.
+constraints:
+  , validation < 1.2
```
carbolymer: Is this still needed? o-c has revision with this constraint:

- https://github.com/IntersectMBO/cardano-haskell-packages/pull/1378
- https://chap.intersectmbo.org/package/ouroboros-consensus-3.0.1.0/

carbolymer (approved):

carbolymer:
Removed from queue, could you double check that validation constraint, @erikd ? Otherwise LGTM.

cabal.project
erikd: But this tree is still using `ouroboros-consensus-3.0.0.0` so that constraint is needed. Unless you would prefer I do a revision of  `ouroboros-consensus-3.0.0.0`

carbolymer: That's strange why. We have only `^>= 3.0` constraint so it should pull `3.0.1` with newer chap I think.

If that is not the case, we can merge your PR as it is.

erikd: I made a revision to `ouroboros-consensus-3.0.0.0` to force `validation < 1.2` and I removed the constraint as well.

PR #1399, opened 2026-07-25
Fix query UTxO rendering

### LINKED ISSUE #1398: [BUG] - query utxo --output-text renders dangling "+ " after value for pre-Babbage outputs
**Summary**

`cardano-cli latest query utxo --output-text` renders a dangling `+ ` separator after the lovelace amount for pre-Babbage era outputs - trailing `+` and space with nothing after it. This is an unannounced change to the text output format in `cardano-cli-11.1.0.0` that breaks scripts parsing this output.

**Steps to reproduce**

On a local testnet that is still in Shelley era (e.g. during a staged hard-fork bootstrap):

```
$ cardano-cli latest query utxo --testnet-magic 42 --output-text --address <addr>
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
8de6d2076c030674c828d4958f4a33d8368579afe8b077d75dfa324fdb6745a1     0        36000000000000000 lovelace +
```

Note the line ends with `lovelace + `.

**Expected behavior**

If there is no datum to show, no separator at all (`... lovelace`), or a placeholder after the `+` as in previous versions. No dangling `+ `.

**Actual behavior**

Pre-Babbage era outputs (Shelley/Allegra/Mary/Alonzo) render as `... lovelace + ` — the datum placeholder is empty, but the separator is still emitted.

Scripts that match on `lovelace$` or `lovelace \+ TxOutDatumNone$` silently select zero UTxOs.

**Version**

`cardano-cli 11.1.0.0 - linux-x86_64 - ghc-9.6` (git rev f098c09d233b021395b123ac77787b9dc4c2595d)

Previous versions (≤ 11.0.x) printed `... lovelace + TxOutDatumNone`.

**Root cause**

Commit f2f7d4d4e ("Remove non-Byron TxOut CtxTx/CtxUTxO usage from cardano-cli", merged in #1379) rewrote `utxoToText` in `cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs` to render from ledger types. The datum is now read via `datumTxOutL`, which only exists for Babbage+ eras, so pre-Babbage outputs render an empty placeholder - but the `" + "` separator is still emitted unconditionally:

```haskell
      , "        " <> printableValue <> " + " <> printableDatum
```

**Suggested fix**

- Suppress the `" + "` separator when the datum placeholder is empty.
- Add a golden test for `query utxo --output-text` so future format changes are caught.

palas:
# Context

`cardano-cli latest query utxo --output-text` renders a dangling `+` separator after the lovelace amount for pre-Babbage era outputs with nothing after it. See https://github.com/IntersectMBO/cardano-cli/issues/1398.

This PR removes the `+` for pre Alonzo eras. And it adds a `NoDatum` tag when it is not present for Alonzo onwards. It also adds the hash for Alonzo, which was ignored before, and golden tests exercising the 3 cases.

# How to trust this PR

The actual changes are very small, and they are very type driven. Also this PR includes golden tests.

# Checklist

- [x] Self-reviewed the diff


cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1225,15 +1225,16 @@ filteredUTxOsToText sbe utxo =
     "                           TxHash                                 TxIx        Amount"
 
 utxoToText
-  :: ShelleyBasedEra era
+  :: forall era
```
Jimbo4350, line 1230: Unnecessary `forall era`.

Jimbo4350 (approved):
LGTM, one minor comment.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
carbolymer: I think it is needed for `ScopedTypeVariables` in line 1265. But then, I think that type annotation might not be necessary.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1246,15 +1248,24 @@ utxoToText sbe (TxIn (TxId txhash) (TxIx index), Exp.TxOut ledgerTxOut) =
...
-      sbe
+    forEraInEon (convert sbe) Nothing $ \aeo ->
+      alonzoEraOnwardsConstraints aeo $
+        Just $
+          forEraInEon
+            (convert sbe)
+            ( Text.pack . show $
+                case ledgerTxOut ^. L.dataHashTxOutL of
+                  L.SNothing -> L.NoDatum :: L.Datum (ShelleyLedgerEra era)
+                  L.SJust dataHash -> L.DatumHash dataHash
```
carbolymer: I'd write it pointfree with `L.strictMaybe` but it's fine as it is 👍🏻

carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
palas: It seems it fails in 9.6 and 9.10 if I remove it:
```
src/Cardano/CLI/EraBased/Query/Run.hs:1264:72: error: [GHC-64725]                                           
      • Cannot satisfy: cardano-ledger-binary-1.8.1.0:Cardano.Ledger.Binary.Version.MinVersion <= L.ProtVerHigh                              
                                                                                                    era0                                     
      • In the second argument of ‘(^.)’, namely ‘L.datumTxOutL’                                                                             
        In the second argument of ‘($)’, namely                                                                                              
          ‘ledgerTxOut ^. L.datumTxOutL’                                                                                                     
        In the expression:                                                                                                                   
          babbageEraOnwardsConstraints beo $ ledgerTxOut ^. L.datumTxOutL                                                                    
       |                                                                                                                                     
  1264 |             (\beo -> babbageEraOnwardsConstraints beo $ ledgerTxOut ^. L.datumTxOutL)                                               
       |                                                                        ^^^^^^^^^^^^^       
```

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/1399/changes/59b2b43fa51a377198adb09f3a9921b931c987a5

PR #1401, opened 2026-07-28
Add conway query gov-action-status command

disassembler:
Implements `cardano-cli conway query gov-action-status --gov-action-id <id>` which queries a live node and displays ratification stage, expiry epoch, DRep vote ratio (stake-weighted, mirroring ledger logic), CC vote ratio (with resignation accounting via JSON introspection), and SPO vote counts.

Example output:
```
Governance action: gov_action1mr0qdz2jmagvsch6r08fhqvq6vu8jakt4c8m9s7ea7z0p740vntqqqqpqyz47
Action type:       TreasuryWithdrawals
Proposed in:       Epoch 639
Expires after:     Epoch 645
Current epoch:     646
Ratify stage:      RATIFIED (pending enactment at epoch end)

=== DRep Votes ===
  Threshold: 67.00%
  YES:       2658457575.26 ADA
  Total:     3449660735.39 ADA (excl. abstain + inactive)
  Ratio:     77.06%
  Status:    PASSING

=== CC Votes ===
  Threshold:      66.67%
  Total in state: 8 (1 resigned, 0 expired)
  Active:         7 (excl. resigned)
  YES votes:      6
  NO votes:       0
  Abstain votes:  0
  Ratio:          85.71%
  Status:         PASSING

=== SPO Votes ===
  Not applicable for TreasuryWithdrawals
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
@@ -88,6 +88,7 @@ import Cardano.Ledger.Conway.Governance
   , RatifyState (..)
   , proposalsActionsMap
   )
+import Cardano.Ledger.Conway.PParams (DRepGroup (..), PPGroups (..), StakePoolGroup (..), modifiedPPGroups)
 import Cardano.Ledger.Conway.State (ChainAccountState (..))
 import Cardano.Slotting.EpochInfo (EpochInfo (..), epochInfoSlotToUTCTime, hoistEpochInfo)
 import Cardano.Slotting.Time (RelativeTime (..), toRelativeTime)
@@ -1619,17 +1620,20 @@ runQueryGovActionStatus
   Cmd.QueryGovActionStatusCmdArgs
     { Cmd.eon
     , Cmd.commons = Cmd.QueryCommons{Cmd.nodeConnInfo, Cmd.target}
-    , Cmd.govActionId = govActionId
+    , Cmd.govActionId
+    , Cmd.outputFormat
+    , Cmd.mOutFile
     } = conwayEraOnwardsConstraints eon $ do
   let sbe = convert eon
-
   govState <- fromExceptTCli $ runQuery nodeConnInfo target $ queryGovState eon
   let bech32Id = Text.unpack (encodeCip129GovernanceActionIdText govActionId)
   gas <- case Map.lookup govActionId (proposalsActionsMap (govState ^. proposalsGovStateL)) of
     Nothing -> do
-      liftIO $ putStrLn $ "Governance action not found in current proposals: " <> bech32Id
-      liftIO $ putStrLn "(Action may have been enacted or expired and removed from the ledger state.)"
-      exitSuccess
+      liftIO $ IO.hPutStrLn IO.stderr $
+        "Governance action not found in current proposals: " <> bech32Id
+      liftIO $ IO.hPutStrLn IO.stderr
+        "(Action may have been enacted or expired and removed from the ledger state.)"
+      liftIO exitFailure
     Just g -> pure g
   committeeState <-
     fromExceptTCli $ runQuery nodeConnInfo target $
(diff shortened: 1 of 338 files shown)


cardano-cli/src/Cardano/CLI/Type/Error/QueryCmdError.hs
```diff
@@ -40,6 +42,7 @@ data QueryCmdError
   | QueryBackwardCompatibleError
       !Text
       !SomeException
+  | QueryCmdGovActionNotFound !L.GovActionId
```
carbolymer: This is not used anywhere.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+  putStrLn $ "Governance action: " <> Text.unpack (encodeCip129GovernanceActionIdText thisGasId)
+  putStrLn $ "Action type:       " <> actionTag
+  putStrLn $ "Proposed in:       Epoch " <> show proposedInN
+  putStrLn $ "Expires after:     Epoch " <> show expiresAfterN
+  putStrLn $ "Current epoch:     " <> show curEpochN
+  if isEnacted
+    then putStrLn   "Ratify stage:      RATIFIED (pending enactment at epoch end)"
+    else if isExpired
+      then putStrLn $ "Ratify stage:      EXPIRED (expired " <> show (curEpochN - expiresAfterN) <> " epoch(s) ago)"
+      else putStrLn   "Ratify stage:      Active (not yet ratified)"
```
carbolymer: rendering of the output should be separate from the computation.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+    } = conwayEraOnwardsConstraints eon $ do
+  let sbe = convert eon
+
+  govState <- fromExceptTCli $ runQuery nodeConnInfo target $ queryGovState eon
+  let bech32Id = Text.unpack (encodeCip129GovernanceActionIdText govActionId)
+  gas <- case Map.lookup govActionId (proposalsActionsMap (govState ^. proposalsGovStateL)) of
+    Nothing -> do
+      liftIO $ putStrLn $ "Governance action not found in current proposals: " <> bech32Id
+      liftIO $ putStrLn "(Action may have been enacted or expired and removed from the ledger state.)"
+      exitSuccess
```
carbolymer: 1. well this is definitely not a success. a different error code should be used.
2. Error messages go to stderr

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+                  Nothing      -> (yes, L.Coin (unCoin tot + stake))
+                  Just L.VoteYes  -> (L.Coin (unCoin yes + stake), L.Coin (unCoin tot + stake))
+                  Just L.VoteNo   -> (yes, L.Coin (unCoin tot + stake))
+                  Just L.Abstain  -> (yes, tot)
+
+showAda :: L.Coin -> String
+showAda (L.Coin lovelace) =
+  let ada = lovelace `div` 1_000_000
+      frac = lovelace `mod` 1_000_000
+  in show ada <> if frac == 0 then "" else "." <> take 2 (show frac)
```
carbolymer: First of all this is unaware of the locale, so it assumes one arbitrary formatting. Secondly, this should be implemented as a newtype wrapper + Pretty instance e.g. `newtype DecimalFormat foo`

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+
+showPct :: Double -> String
+showPct x =
+  let pct = round (x * 10_000) :: Integer
+      whole = pct `div` 100
+      frac  = pct `mod` 100
+      pad2 n = if n < 10 then "0" <> show n else show n
+  in show whole <> "." <> pad2 frac <> "%"
```
carbolymer: same as above

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+  putStrLn $ "Action type:       " <> actionTag
+  putStrLn $ "Proposed in:       Epoch " <> show proposedInN
+  putStrLn $ "Expires after:     Epoch " <> show expiresAfterN
+  putStrLn $ "Current epoch:     " <> show curEpochN
+  if isEnacted
+    then putStrLn   "Ratify stage:      RATIFIED (pending enactment at epoch end)"
+    else if isExpired
+      then putStrLn $ "Ratify stage:      EXPIRED (expired " <> show (curEpochN - expiresAfterN) <> " epoch(s) ago)"
+      else putStrLn   "Ratify stage:      Active (not yet ratified)"
+  putStrLn ""
```
carbolymer: It should support different output formats e.g. json / yaml similarly to other commands.

carbolymer (changes requested):
I suppose the feature could be useful 👍🏻 But the code needs more work.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+drepThresholdFor :: String -> L.DRepVotingThresholds -> Maybe Double
+drepThresholdFor tag L.DRepVotingThresholds{..} = fmap (fromRational . L.unboundRational) $ case tag of
+  "TreasuryWithdrawals" -> Just dvtTreasuryWithdrawal
+  "HardForkInitiation"  -> Just dvtHardForkInitiation
+  "NoConfidence"        -> Just dvtMotionNoConfidence
+  "UpdateCommittee"     -> Just dvtCommitteeNormal
+  "NewConstitution"     -> Just dvtUpdateToConstitution
+  "ParameterChange"     -> Just dvtPPGovGroup
+  "InfoAction"          -> Nothing
+  _                     -> Nothing
```
carbolymer: **ParameterChange thresholds diverge from ledger ratification logic.**

`drepThresholdFor "ParameterChange" = Just dvtPPGovGroup` hardcodes a single group, but the ledger (`votingDRepThresholdInternal` in `Cardano.Ledger.Conway.Governance.Internal`) computes the threshold as the **maximum over all PParam groups the proposal actually touches** (Network / Economic / Technical / Gov). A proposal changing, say, an economic and a technical parameter will show the wrong threshold here.

Similarly, `spoRequired "ParameterChange" = False` unconditionally, but the ledger requires SPO votes (at `pvtPPSecurityGroup`) whenever the proposal touches a **SecurityGroup** parameter (fee params, max block/tx sizes, coins-per-UTxO-byte, ...). As written, the command omits the SPO section exactly for the security-relevant parameter changes where it matters most.

Both need to inspect the `ParameterChange` payload's `PParamsUpdate` and derive the touched groups, as the ledger does, rather than dispatching on the action tag alone.

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+          resignedCount = Map.size $ Map.filter (\v -> isActive v && isResigned v) memberJsons
+          expiredCount  = Map.size $ Map.filter (\v -> getStatus v == Just "Expired") memberJsons
+          totalInState  = Map.size csCommittee
+          activeCount   = Map.size activeNR
+          yesVotes      = Map.size $ Map.filter (== L.VoteYes) gasCommitteeVotes
+          noVotes       = Map.size $ Map.filter (== L.VoteNo) gasCommitteeVotes
+          abstainVotes  = Map.size $ Map.filter (== L.Abstain) gasCommitteeVotes
+          ccRatio       = if activeCount == 0 then 0.0 else fromIntegral yesVotes / fromIntegral activeCount :: Double
+          ccThreshD     = maybe 0.0 (fromRational . L.unboundRational) csThreshold
+          ccNeeded      = max 0 (ceiling (ccThreshD * fromIntegral activeCount) - yesVotes) :: Int
```
carbolymer: **CC ratio denominator disagrees with the ledger: explicit Abstain votes must be excluded.**

`activeCount` counts every active, non-resigned member regardless of how they voted, but the ledger's `committeeAcceptedRatio` (`Cardano.Ledger.Conway.Rules.Ratify`) removes members who explicitly voted `Abstain` from the denominator as well (same treatment as resigned/expired/unregistered).

With the current code the reported ratio is systematically lower than the ledger's whenever anyone abstains, so the command can print `NOT PASSING` for an action the ledger will in fact ratify. Example: 7 active members, threshold 66.67%, 4 yes, 1 abstain - ledger ratio 4/6 = 66.7% (passing), this code 4/7 = 57.1% (reported not passing).

cardano-cli/src/Cardano/CLI/EraBased/Query/Run.hs
```diff
@@ -1598,6 +1612,242 @@ runQueryConstitution
...
+  "NewConstitution"     -> Just dvtUpdateToConstitution
+  "ParameterChange"     -> Just dvtPPGovGroup
+  "InfoAction"          -> Nothing
+  _                     -> Nothing
+
+spoThresholdFor :: String -> L.PoolVotingThresholds -> Maybe Double
+spoThresholdFor tag L.PoolVotingThresholds{..} = fmap (fromRational . L.unboundRational) $ case tag of
+  "NoConfidence"       -> Just pvtMotionNoConfidence
+  "HardForkInitiation" -> Just pvtHardForkInitiation
+  "UpdateCommittee"    -> Just pvtCommitteeNormal
```
carbolymer: **`UpdateCommittee` always uses the `*Normal` thresholds; the ledger switches to `*NoConfidence` when no committee is seated.**

`dvtCommitteeNormal` (above) and `pvtCommitteeNormal` (here) are only correct while a committee exists. After a successful NoConfidence action (state of no confidence), the ledger uses `dvtCommitteeNoConfidence` / `pvtCommitteeNoConfidence` instead. The committee presence is available from the gov/ratify state already being queried, so this can be dispatched on rather than hardcoded.

carbolymer (commented):
Three correctness observations on the vote-ratio logic, verified against the ledger's ratification code (`Cardano.Ledger.Conway.Rules.Ratify` / `Governance.Internal`). The DRep AlwaysAbstain/AlwaysNoConfidence/expiry handling in `calcDRepRatio` itself matches the ledger's `dRepAcceptedRatio` - the issues below are in the threshold selection and the CC denominator.

PR #1405, opened 2026-08-04
Guard against network id mismatches

### LINKED ISSUE #1402: query stake-address-info: mainnet address rendered with stake_test1 prefix in output
## Description

`cardano-cli query stake-address-info` renders the `address` field in its output using the `stake_test1` bech32 prefix even when querying a mainnet address.

## Steps to reproduce

```
cardano-cli query stake-address-info \
  --mainnet \
  --address stake1u85was6qks0exltkfl74clay4p6l272ry6kpkgweyfs7vtcvaj7xm
```

## Actual output

```json
[
    {
        "address": "stake_test1ur5was6qks0exltkfl74clay4p6l272ry6kpkgweyfs7vtcthcuzx",
        ...
    }
]
```

The `address` field uses the `stake_test1` prefix, which is incorrect for a mainnet address.

## Expected output

The `address` field should use the `stake1` prefix, matching the network of the query:

```json
[
    {
        "address": "stake1u85was6qks0exltkfl74clay4p6l272ry6kpkgweyfs7vtcvaj7xm",
        ...
    }
]
```

## Notes

The input address (`stake1u...`) is correctly recognised as a mainnet address and the query succeeds. Only the rendered output address uses the wrong prefix. The underlying stake credential is identical in both cases.

palas:
# Context

This PR aims to address https://github.com/IntersectMBO/cardano-cli/issues/1402. I could find an explanation for why would the CLI would use the testnet bech32 prefix for the response to the `cardano-cli query stake-address-info` command when using the `--mainnet` flag.

But I can see this happening if `--testnet 764824073` is specified when it is actually mainnet, or if `CARDANO_NODE_NETWORK_ID=764824073` as env variable instead of `CARDANO_NODE_NETWORK_ID=mainnet`. And it shouldn't. Also the mainnet/testnet from the address input is discarded.

This PR ensures the node now checks that the network id given to the CLI (`--mainnet`, `--testnet-magic`, or `CARDANO_NODE_NETWORK_ID`) matches the network id in the node's genesis, so that for example `CARDANO_NODE_NETWORK_ID=764824073` against a mainnet node no longer renders addresses with a testnet prefix. 

Additionally, `query stake-address-info` now fails when the given stake address does not match the network id, instead of silently dropping the address's network tag.

# How to trust this PR

Changes are small. The `stake-address-info` query change has a test. And the other change can be tested against a node, but making a test for it in `cardano-cli` is tricky.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
+++ b/.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
@@ -1,10 +1,4 @@
-description: Commands that query a node now check that the network id given to
-  the CLI (`--mainnet`, `--testnet-magic`, or `CARDANO_NODE_NETWORK_ID`) matches
-  the network id in the node's genesis, so that for example
-  `CARDANO_NODE_NETWORK_ID=764824073` against a mainnet node no longer renders
-  addresses with a testnet prefix. Additionally, `query stake-address-info` now
-  fails when the given stake address does not match the network id, instead of
-  silently dropping the address's network tag.
+description: Commands that query a node now check that the network id given to the CLI (`--mainnet`, `--testnet-magic`, or `CARDANO_NODE_NETWORK_ID`) matches the network id in the node's genesis, so that for example `CARDANO_NODE_NETWORK_ID=764824073` against a mainnet node no longer renders addresses with a testnet prefix. Additionally, `query stake-address-info` now fails when the given stake address does not match the network id, instead of silently dropping the address's network tag.
 kind:
 - bugfix
 - test
--- a/.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
+++ b/.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
@@ -1,0 +1,12 @@
+description: Commands that query a node now check that the network id given to
+  the CLI (`--mainnet`, `--testnet-magic`, or `CARDANO_NODE_NETWORK_ID`) matches
+  the network id in the node's genesis, so that for example
+  `CARDANO_NODE_NETWORK_ID=764824073` against a mainnet node no longer renders
+  addresses with a testnet prefix. Additionally, `query stake-address-info` now
+  fails when the given stake address does not match the network id, instead of
+  silently dropping the address's network tag.
+kind:
+- bugfix
+- test
+pr: 1405
+project: cardano-cli
--- a/cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
+++ b/cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
@@ -1,7 +1,7 @@
-{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE TypeApplications #-}
 
 module Cardano.CLI.LocalStateQuery
-  ( executeLocalStateQueryExprWithNetworkIdCheck
+  ( checkNodeNetworkId
   )
 where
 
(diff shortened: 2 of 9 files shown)


.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
```diff
@@ -0,0 +1,12 @@
+description: Commands that query a node now check that the network id given to
+  the CLI (`--mainnet`, `--testnet-magic`, or `CARDANO_NODE_NETWORK_ID`) matches
+  the network id in the node's genesis, so that for example
+  `CARDANO_NODE_NETWORK_ID=764824073` against a mainnet node no longer renders
+  addresses with a testnet prefix. Additionally, `query stake-address-info` now
+  fails when the given stake address does not match the network id, instead of
+  silently dropping the address's network tag.
+kind:
```
carbolymer: redundant line breaks

cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
```diff
@@ -0,0 +1,58 @@
...
+          throwCliError $ NodeNetworkIdMismatchError cliNetId nodeNetId
+    _ -> pure ()
+
+-- | The network id from the node's genesis, or 'Nothing' when it cannot be
+-- obtained (the node is still in the Byron era, or it does not support the
+-- necessary queries): the absence of an answer is not treated as a mismatch.
+queryNodeNetworkId
+  :: LocalStateQueryExpr BlockInMode ChainPoint QueryInMode () IO (Maybe NetworkId)
+queryNodeNetworkId =
+  queryCurrentEra >>= \case
```
carbolymer: suggestion, use Monad in `do` instead of manually pattern matching in a staircase

cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
```diff
@@ -0,0 +1,58 @@
...
+-- with the right magic but the wrong tag (for example
+-- @CARDANO_NODE_NETWORK_ID=764824073@ instead of @CARDANO_NODE_NETWORK_ID=mainnet@)
+-- connects successfully and would otherwise make the CLI render addresses for
+-- the wrong network.
+executeLocalStateQueryExprWithNetworkIdCheck
+  :: LocalNodeConnectInfo
+  -> Consensus.Target ChainPoint
+  -> LocalStateQueryExpr BlockInMode ChainPoint QueryInMode () IO a
+  -> IO (Either AcquiringFailure a)
+executeLocalStateQueryExprWithNetworkIdCheck connectInfo target f =
```
carbolymer: This feels excessive. I think it may be more readable to just put explicit `checkNodeNetwork` in those few places where it's needed rather than creating a specialised query execution wrapper.

.changes/20260804_011228_cardano-cli_pablo.lamela_check_network_id.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/1405/changes/af04b29e6f1a3dd6c1d4cbb14a65dbed72f5f035

cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/1405/changes/af04b29e6f1a3dd6c1d4cbb14a65dbed72f5f035

cardano-cli/src/Cardano/CLI/LocalStateQuery.hs
palas: I think it is something we must do for every local state query, in the same way the handshake checks the magic number, so having the wrapper makes it more likely that new queries will also do this. And having the wrapper is less code than adding it everywhere, cause there are more than more than 20 places

Jimbo4350: I agree and I've thought more about it. We should have a single query that is run once before the relevant command. Not on every single query. Or what would be better is if the check ran before we `case`d in `runQueryCmds`.

Jimbo4350 (changes requested):

palas:
> I agree and I've thought more about it. We should have a single query that is run once before the relevant command. Not on every single query. Or what would be better is if the check ran before we `case`d in `runQueryCmds`.

Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/1405/changes/d1e3ee982ddf39f8aa450e366faaa9785011bb86

Jimbo4350 (approved):
LGTM

PR #1423, opened 2026-08-21
Register a stake pool's BLS voting key

ch1bo:
Adds `--bls-signing-key-file` to `stake-pool registration-certificate`.
The option is era-gated: mandatory from Dijkstra onwards and not offered
at all in Conway or earlier, so existing invocations are unaffected. The
signing key is read as a text envelope and turned into the registered
verification key plus proof of possession via `createBlsKeyRegistration`.

Genesis pool parameters get `sppBlsKey = SNothing`, which
`cardano-ledger-core 1.22` requires at every `StakePoolParams` construction
site, and the compatible pre-Dijkstra path sets `stakePoolBlsKey = Nothing`.

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


.changes/20260904_fix_dijkstra_era_commands.yml
```diff
@@ -0,0 +1,13 @@
+project: cardano-cli
+
+pr: 1439
+
+kind:
+  - bugfix
+
+description: |
+  Fix Dijkstra era commands (BLS key generation/hashing, proof of possession, transaction
```
carbolymer, line 9: line wraps need to go

PR #1427, opened 2026-08-27
Support Dijkstra simple scripts read from JSON

palas:
# Context

Reading a simple script from a JSON file in the Dijkstra era aborted with an internal `TODO Dijkstra` error. The reader parses simple scripts in the shared Allegra timelock format and then converts to the era's script type; the Dijkstra arm was a stub.

This PR fills it using `upgradeTimelock`, the same conversion the api uses for this exact purpose.

Today this path is not reachable from the command line: no Dijkstra command takes script files yet. That arrives in a later PR, and this stub has to be gone before it does, otherwise every Dijkstra command reading a simple script would crash.

# How to trust this PR

Pretty much the same mechanism as for Conway, but using `Dijkstra.upgradeTimelock`.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/EraBased/Script/Read/Common.hs
+++ b/cardano-cli/src/Cardano/CLI/EraBased/Script/Read/Common.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
 
 module Cardano.CLI.EraBased.Script.Read.Common
   ( -- * Plutus Script Related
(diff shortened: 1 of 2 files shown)


cardano-cli/src/Cardano/CLI/EraBased/Script/Read/Common.hs
```diff
@@ -39,10 +41,16 @@ readFileSimpleScript file era = do
...
+        script :: SimpleScript <- fromEitherCli $ Aeson.eitherDecodeStrict' bs
+        let s :: L.NativeScript (Exp.LedgerEra era) =
+              Dijkstra.upgradeTimelock (toAllegraTimelock @(Exp.LedgerEra Exp.ConwayEra) script)
+        return $ Exp.SimpleScript s
       Exp.ConwayEra -> Exp.obtainConwayConstraints era $ do
         -- In addition to the TextEnvelope format, we also try to
-        -- deserialize the JSON representation of SimpleScripts..
+        -- deserialize the JSON representation of SimpleScripts.
         script :: SimpleScript <- fromEitherCli $ Aeson.eitherDecodeStrict' bs
         let s :: L.NativeScript (Exp.LedgerEra era) = obtainCommonConstraints era $ toAllegraTimelock script
```
carbolymer: I think this can be refactored outside of the case:
```haskell
allegraTimelock :: L.NativeScript (Exp.LedgerEra era) <- Exp.obtainConwayConstraints era $ (toAllegraTimelock <$> fromEitherCli (Aeson.eitherDecodeStrict' bs))
```
it'd make `case` smaller:

```haskell
pure . Exp.SimpleScript $ case era of
 Exp.DijkstraEra -> Dijkstra.upgradeTimelock allegraTimelock
 Exp.ConwayEra -> allegraTimelock
```
btw. `@(Exp.LedgerEra Exp.ConwayEra)` is not needed because `toAllegraTimelock` is well typed.

carbolymer (approved):

cardano-cli/src/Cardano/CLI/EraBased/Script/Read/Common.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-cli/pull/1427/changes/f8281e5f7ee003e2924b16ff84b520ceaf754aaa

Jimbo4350 (approved):

PR #1430, opened 2026-08-27
Render Dijkstra certificates in the friendly JSON view

palas:
# Context

The friendly JSON renderer (what `transaction view` prints) hit an error stub for every Dijkstra-era certificate.

This PR adds `renderDijkstraCertificate`: it is the Conway renderer minus the certificate forms the era drops (Dijkstra no longer has the Shelley-era certificate shapes), so every certificate the era supports renders, and stake pool registrations include their pool parameters.

# How to trust this PR

It is all straightforward to JSON conversion. The only doubt is whether we are covering all the right certificates (since there is a catch-all that prints `"unsupported certificate"`. But it can be checked that they match the ones at: https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-dijkstra/Cardano-Ledger-Dijkstra-TxCert.html#t:DijkstraTxCert

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
+++ b/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
@@ -766,7 +766,7 @@ renderDijkstraCertificate cert =
       "Drep certificate update"
         .= object
           [ "Drep credential" .= drepCredential
-          , "anchor " .= mbAnchor
+          , "anchor" .= mbAnchor
           ]
     _ -> "unsupported certificate" .= String (T.pack $ show cert)
 
--- a/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
+++ b/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
@@ -554,7 +554,7 @@ renderCertificate
 renderCertificate era (Exp.Certificate c) =
   case era of
     Exp.ConwayEra -> renderConwayCertificate c
-    Exp.DijkstraEra -> error "renderCertificate: TODO Dijkstra era not supported"
+    Exp.DijkstraEra -> renderDijkstraCertificate c
 
 renderDrepCredential
   :: ()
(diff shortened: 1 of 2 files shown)


cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
```diff
@@ -681,6 +681,95 @@ renderConwayCertificate cert =
...
+          [ "stake pool key hash" .= kh
+          , "epoch" .= epoch
+          ]
+    L.UpdateDRepTxCert drepCredential mbAnchor ->
+      "Drep certificate update"
+        .= object
+          [ "Drep credential" .= drepCredential
+          , "anchor " .= mbAnchor
+          ]
+    _ -> "unsupported certificate" .= String (T.pack $ show cert)
```
carbolymer, line 771: what's hidden in `_`? Shouldn't that be an error? at least a non-zero error code. This applies to conway as well.
I don't remember why `COMPLETE` pragma doesn't work here.

carbolymer (approved):

cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
palas: There is a `COMPLETE` pragma here: https://github.com/IntersectMBO/cardano-ledger/blob/f3104f00f9819ba94de119c38bc3e0109982821f/eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs#L358-L371

In the first place, it doesn't work because it is done over `DijkstraEra` instead of `DijkstraTxCert`, but that also doesn't seem to work, may be a compiler issue.

So the only way to get it to work would be to recursively pattern match on [DijkstraTxCert](https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-dijkstra/Cardano-Ledger-Dijkstra-TxCert.html#t:DijkstraTxCert), but we are not supposed to do that, and is not wat we did for Conway.

And about it being an error instead: I think it is best to just print it uglily than failing if for some reason we miss a certificate

cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs
```diff
@@ -681,6 +681,95 @@ renderConwayCertificate cert =
...
+        .= object
+          [ "stake pool key hash" .= kh
+          , "epoch" .= epoch
+          ]
+    L.UpdateDRepTxCert drepCredential mbAnchor ->
+      "Drep certificate update"
+        .= object
+          [ "Drep credential" .= drepCredential
+          , "anchor " .= mbAnchor
+          ]
```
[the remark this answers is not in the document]
palas: The trailing space is fixed in: https://github.com/IntersectMBO/cardano-cli/pull/1430/changes/a1c0428055bb337f8d9740f2e609365aca075849
The inconsistency is not true, it is done the same in Conway: https://github.com/IntersectMBO/cardano-cli/blob/eefdd5c0a70f88c768371c6860cac8c526161d71/cardano-cli/src/Cardano/CLI/Compatible/Json/Friendly.hs#L679

Jimbo4350 (approved):

Issue #1422, opened 2026-08-20
`transaction assemble` cannot read the witnesses that `transaction witness` writes in Dijkstra


mkoura:
### Summary

`cardano-cli transaction witness` writes a text envelope of type
`TxWitness DijkstraEra`, but `cardano-cli transaction assemble` (alias
`sign-witness`) only accepts `TxWitness ShelleyEra` through
`TxWitness ConwayEra`. The witness round-trip is broken inside a single binary
whenever the era is Dijkstra.

### Steps to reproduce

Against a Dijkstra testnet (protocol version 12), with cardano-cli 11.1.0.0:

```sh
cardano-cli dijkstra transaction build-raw ... --out-file tx.body
cardano-cli dijkstra transaction witness --tx-body-file tx.body --signing-key-file payment.skey --out-file tx.witness
cardano-cli dijkstra transaction assemble --tx-body-file tx.body --witness-file tx.witness --out-file tx.signed
```

### Actual

```
Error: Failed to decode the ledger's CDDL serialisation format. TextEnvelope type error:  Expected one of: TxWitness ShelleyEra, TxWitness AllegraEra, TxWitness MaryEra, TxWitness AlonzoEra, TxWitness BabbageEra, TxWitness ConwayEra Actual: TxWitness DijkstraEra
CallStack (from HasCallStack):
  fromExceptTCli, called at src/Cardano/CLI/EraBased/Transaction/Run.hs:119:41 in cardano-cli-11.1.0.0:Cardano.CLI.EraBased.Transaction.Run
  runTransactionCmds, called at src/Cardano/CLI/EraBased/Run.hs:61:5 in cardano-cli-11.1.0.0:Cardano.CLI.EraBased.Run
  runCmds, called at src/Cardano/CLI/EraBased/Run.hs:35:35 in cardano-cli-11.1.0.0:Cardano.CLI.EraBased.Run
  runAnyEraCommand, called at src/Cardano/CLI/Run.hs:56:5 in cardano-cli-11.1.0.0:Cardano.CLI.Run
  runClientCommand, called at app/cardano-cli.hs:58:14 in cardano-cli-11.1.0.0:Main
```

with `tx.witness` containing exactly what the previous command wrote:

```json
{
    "type": "TxWitness DijkstraEra",
    "description": "Key Witness ShelleyEra",
    "cborHex": "8258..."
}
```

### Expected

`transaction assemble` accepts a witness produced by `transaction witness` of
the same binary and era.

[...]

PR #1432, opened 2026-08-28
Add an end-to-end test for Dijkstra transaction witnessing

palas:
# Context

This is the regression test for #1422: `transaction assemble` used to reject witness
files from eras after Conway, and the fix merged without its test because the test needs
`dijkstra transaction` commands to exist on master, which only happened in the PR below
this one.

# How to trust this PR

- Test-only: zero changes to library or executable code
- The test passes
- It is a simple workflow

# Checklist

- [x] Self-reviewed the diff


cardano-cli/test/cardano-cli-golden/Test/Golden/Dijkstra/Transaction/Assemble.hs
```diff
@@ -0,0 +1,73 @@
...
+        , "assemble"
+        , "--tx-body-file"
+        , txBodyFile
+        , "--witness-file"
+        , witnessFile
+        , "--out-file"
+        , signedTxFile
+        ]
+
+    H.assertFileOccurences 1 "Tx DijkstraEra" signedTxFile
```
carbolymer: I think this is a weak test. Conway transaction assemble test, tests against a golden file (I presume that was tested that the transaction is fine?).

palas: You are right. It was a regression test for https://github.com/IntersectMBO/cardano-cli/issues/1422, but why not have a golden file. Addressed it in: https://github.com/IntersectMBO/cardano-cli/pull/1432/changes/4e99a781f3192da89a07bd5cec93f4e699dfac7d

cardano-cli/test/cardano-cli-golden/Test/Golden/Dijkstra/Transaction/Assemble.hs
```diff
@@ -0,0 +1,76 @@
...
+        , "--tx-body-file"
+        , txBodyFile
+        , "--witness-file"
+        , witnessFile
+        , "--out-file"
+        , signedTxFile
+        ]
+
+    goldenSignedTxFile <-
+      H.note "test/cardano-cli-golden/files/golden/dijkstra/transaction/assemble_out"
```
carbolymer, line 79: do we know if that transaction is ok - is accepted by ledger?

carbolymer (approved):

cardano-cli/test/cardano-cli-golden/Test/Golden/Dijkstra/Transaction/Assemble.hs
palas: I haven't tested that transaction in particular, I'll try spawning a dijkstra testnet and getting a manually tested transaction here

palas: Addressed in: https://github.com/IntersectMBO/cardano-cli/pull/1432/changes/62e6b19e3166c92581e8eabb20b4123309440688

PR #1438, opened 2026-09-03
Replace caseShelleyToBabbageOrConwayEraOnwards with inEonForShelleyBasedEra

Jimbo4350:
# Context

[IntersectMBO/cardano-api#1326](https://github.com/IntersectMBO/cardano-api/pull/1326) removes the era case combinators, including `caseShelleyToBabbageOrConwayEraOnwards`. This migrates the call sites here ahead of that.

Most sites ignored one or both of the witnesses that combinator hands out, so they become `inEonForShelleyBasedEra`.

`pUpdateProtocolParametersCmd` was the awkward one: it matched on all seven `ShelleyBasedEra` constructors, but only to build a `ShelleyToBabbageEra` for the `eon` field of `UpdateProtocolParametersPreConway` — a field nothing ever read. Dropping the field removes the match. Its two branches then differ only in which optional payload they fill, so the shared scaffolding moves into `mkCmd`.

`inEonForShelleyBasedEra` is already in the released cardano-api, so this can merge before or after the cardano-api PR.

# How to trust this PR

No change to the CLI's interface — same commands, same help output, which the golden tests cover.

```bash
cabal build cardano-cli -j4
cabal test cardano-cli-golden cardano-cli-test -j4
```

Locally: clean build, `cardano-cli-golden` 815/815, `cardano-cli-test` 73/73, no new `hlint` hints.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Compatible/Governance/Option.hs
+++ b/cardano-cli/src/Cardano/CLI/Compatible/Governance/Option.hs
@@ -1,10 +1,11 @@
+{-# LANGUAGE GADTs #-}
+
 module Cardano.CLI.Compatible.Governance.Option
   ( pCompatibleGovernanceCmds
   )
 where
 
 import Cardano.Api
-import Cardano.Api.Experimental qualified as Exp
 
 import Cardano.CLI.Compatible.Governance.Command
 import Cardano.CLI.Compatible.Governance.Types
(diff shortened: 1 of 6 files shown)


cardano-cli/src/Cardano/CLI/Compatible/Governance/Option.hs
```diff
@@ -33,22 +29,21 @@ pCompatibleGovernanceCmds
...
-                ( Opt.progDesc $
-                    mconcat
-                      [ "Governance commands."
-                      ]
-                )
-                [ pCreateMirCertificatesCmds sbe
-                , pGovernanceGenesisKeyDelegationCertificate
-                , fmap CreateCompatibleProtocolParametersUpdateCmd <$> pGovernanceActionCmds sbe
-                ]
+      [ inEonForShelleyBasedEra
```
palas: Because of the use of `inEonForShelleyBasedEra`, when `Conway` goes out of `Era`, then it will lose the conway related commands (like `drep`), and we won't get any compilation error that warns about it here.
So maybe we should do something like `postConway` below.

palas (approved):
I am pretty confident this is a pure refactoring (from the point of view of the client). Just a minor comment/suggestion

PR #1447, opened 2026-09-18
Remove the stale x86_64-darwin CI badge from the README

carbolymer:
# Context

Removes the `x86_64-darwin` CI badge from the README, because that platform is no longer built.

`supportedSystems` in `flake.nix` lists only `x86_64-linux`, `aarch64-linux` and `aarch64-darwin`, so nothing schedules `x86_64-darwin` any more.
Its Hydra aggregate still answers with a green shield, but it is frozen: the last finished build of `x86_64-darwin.required` is from 2026-05-04, while `x86_64-linux.required` and `aarch64-darwin.required` rebuild on every push.
A permanently green badge for a platform we do not build is worse than no badge.

The remaining two badges cover what Hydra actually builds.
There is deliberately no `aarch64-linux` badge: its Hydra jobs are disabled in `flake.nix` ("we don't have any native builders for this architecture as of 2024-07-15"), so `aarch64-linux.required/shield` returns 404 and would render as a broken badge.
aarch64-linux and Windows binaries are produced by cross-compilation on x86_64-linux, so those builds are already covered by the `x86_64-linux` badge.

# How to trust this PR

Compare the shields before and after:

```
curl -s https://ci.iog.io/job/IntersectMBO-cardano-cli/master/x86_64-darwin.required/latest-finished \
  -H 'Accept: application/json' | jq '.stoptime | todate'
```

That prints a timestamp from May 2026; the same query for `x86_64-linux.required` and `aarch64-darwin.required` prints today's builds.

# Checklist

- [ ] Self-reviewed the diff


palas (approved):
Yes, it hasn't been built since May apparently

